PR #79, opened 2022-07-05
update readme

nikolas-burkoff:
Part of https://github.com/insightsengineering/teal/issues/668

Test with `remotes::install_github("insightsengineering/teal.data@668_instll_td@main")` once https://github.com/insightsengineering/idr-tasks/issues/343 has been completed though note of course that it will install the released version of teal.logger

As we are only using this to install released packages (and staged.deps for when we as devs are developing) this should be all we need)


junlueZH (approved):
LGTM

PR #91, opened 2022-07-26
update vignettes

mhallal1:
part of https://github.com/insightsengineering/teal/issues/700

1) added `join_keys` and `preprocessing data` vignettes from `teal`
2) Replaced instances as `[XXX()]` with `XXX`
3) Removed the long printed datasets when possible and the unnecessary messages


vignettes/preprocessing-data.Rmd
```diff
@@ -0,0 +1,143 @@
...
+author: "NEST CoreDev"
+date: "2022-07-26"
+output: rmarkdown::html_vignette
+vignette: >
+  %\VignetteIndexEntry{Pre-processing data}
+  %\VignetteEngine{knitr::rmarkdown}
+  %\VignetteEncoding{UTF-8}
+---
+
+"Data" preprocessing refers to the code which precedes the `teal` app initialization in `app.R`, i.e.:
```
nikolas-burkoff, line 12: We should probably tweak this vignette as it refers to teal apps but we're in teal.data (we have an eval FALSE on line 21)? Other lines w.g. line 17 don't make sense in the teal.data context

vignettes/join-keys.Rmd
```diff
@@ -0,0 +1,92 @@
...
+Refer to `vignette("including-adam-data-in-teal", package = "teal")` for more information.
+
+### Primary key
+
+Using the `keys` argument to the `dataset` (or for `DDL` a `dataset_connector`) function, we can specify column(s) of the
+dataset which together uniquely identify rows in the dataset. `teal` throws an error if it finds at least two rows
+sharing all the same values of the primary key columns.
+
+```{r, eval=FALSE}
+library(teal)
```
nikolas-burkoff, line 33: I think we should tweak these to not use teal and some of the text may need tweaking now this has been moved

nikolas-burkoff (commented):
- Delayed Data Loading vignette is for people wanting to create their own Connectors etc. and goes into crazy detail whereas Delayed Data Objects vignette is for people wanting to use DDL so maybe we should rename them and add that emphasis at the beginning of the vignettes?

vignettes/preprocessing-data.Rmd
mhallal1: Updated.

vignettes/join-keys.Rmd
mhallal1: Updated.

nikolas-burkoff (approved):

Issue #68, opened 2022-06-03
two print outputs when printing `TealDataset` object


pawelru:
https://github.com/insightsengineering/teal.data/blob/4588016464044cfca5cbd8df75cd525c035cdc28/R/TealDataset.R#L126-L130

Above is printing twice in most of the cases. As an example - please open pkgdown documentation for `cdisc_dataset` - in the examples section you would see two print outputs. 

```
library(scda)

ADSL <- synthetic_cdisc_data("latest")$adsl

cdisc_dataset("ADSL", ADSL, metadata = list(type = "scda", date = "latest"))
#> A CDISCTealDataset object containing the following data.frame (400 rows and 44 columns):
#>   STUDYID               USUBJID SUBJID SITEID AGE  AGEU SEX
#> 1 AB12345  AB12345-CHN-3-id-128 id-128  CHN-3  32 YEARS   M
#> 2 AB12345 AB12345-CHN-15-id-262 id-262 CHN-15  35 YEARS   M
#> 3 AB12345  AB12345-RUS-3-id-378 id-378  RUS-3  30 YEARS   F
#> 4 AB12345 AB12345-CHN-11-id-220 id-220 CHN-11  26 YEARS   F
#> 5 AB12345  AB12345-CHN-7-id-267 id-267  CHN-7  40 YEARS   M
#> 6 AB12345 AB12345-CHN-15-id-201 id-201 CHN-15  49 YEARS   M
#>                        RACE                 ETHNIC COUNTRY DTHFL         INVID
#> 1                     ASIAN NOT HISPANIC OR LATINO     CHN     N  INV ID CHN-3
#> 2 BLACK OR AFRICAN AMERICAN NOT HISPANIC OR LATINO     CHN     N INV ID CHN-15
#> 3                     ASIAN NOT HISPANIC OR LATINO     RUS     N  INV ID RUS-3
#> 4                     ASIAN NOT HISPANIC OR LATINO     CHN     N INV ID CHN-11
#> 5                     ASIAN                UNKNOWN     CHN     N  INV ID CHN-7
#> 6                     ASIAN NOT HISPANIC OR LATINO     CHN     N INV ID CHN-15
#>           INVNAM            ARM ARMCD         ACTARM ACTARMCD         TRT01P
#> 1  Dr. CHN-3 Doe      A: Drug X ARM A      A: Drug X    ARM A      A: Drug X
[...]

PR #116, opened 2022-11-07
two print outputs when printing TealDataset

chlebowa:
Closes [this issue](https://github.com/insightsengineering/teal.data/issues/68).

Slight modification to the `print` method of `TealDataset`.


nikolas-burkoff (approved):
Yup this looks good 👍

PR #119, opened 2022-11-09
remove hashing

mhallal1:
related to https://github.com/insightsengineering/teal/issues/751

remove datasets hashing and transfer it to `teal`.


nikolas-burkoff (approved):
Just NEWS to add here then we're good

nikolas-burkoff:
Actually can we remove digest from teal.data's dependencies as well?

mhallal1:
> Actually can we remove digest from teal.data's dependencies as well?

It is still used in `codeClass.R`

Issue #65, opened 2022-05-30
Long datasets in vignettes


mhallal1:
Datasets are rendered in the `pkgdown` site and they tend to be very long and therefore not user friendly.

PR #130, opened 2022-12-20
fix documentation #65

chlebowa:
+ separated help pages for `TealData` and `MAETealData`
+ added `@seealso` fields to link to each other
+ bonus: fixed the Python vignette to not crash without the `reticulate` package


DESCRIPTION
```diff
@@ -59,4 +59,4 @@ Encoding: UTF-8
 Language: en-US
 LazyData: true
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.2.1
+RoxygenNote: 7.2.2
```
asbates, line 62: Looks like GH Actions [doesn't like this version](https://github.com/insightsengineering/teal.data/actions/runs/3740976714/jobs/6349987680).

R/MAETealDataset.R
```diff
@@ -7,37 +7,33 @@
...
-#'
+#'        A vector of primary keys
 #' @param code (`character` or `CodeClass`)\cr
-#'   A character string defining the code needed to produce the data set in `x`.
-#'   `initialize()` and `recreate()` accept code as `CodeClass` also
-#'   which is needed to preserve the code uniqueness and correct order.
-#'
+#'        A character string defining the code needed to produce the data set in `x`.
+#'        initialize()` and `recreate()` accept code as `CodeClass` also
+#'        which is needed to preserve the code uniqueness and correct order.
```
mhallal1, line 19:
```suggestion
#'        initialize()` and `recreate()` accept code as `CodeClass` which is
#'        also needed to preserve the code uniqueness and correct order.
```

R/MAETealDataset.R
```diff
@@ -7,37 +7,33 @@
...
-#'   It's recommended to include `TealDataset` or `TealDatasetConnector` objects to
-#'   the `vars` list to preserve reproducibility. Please note that `vars`
-#'   are included to this object as local `vars` and they cannot be modified
-#'   within another dataset.
-#'
+#'        In case when this object code depends on other `TealDataset` object(s) or
+#'        other constant value, this/these object(s) should be included as named
+#'        element(s) of the list. For example if this object code needs `ADSL`
+#'        object we should specify `vars = list(ADSL = <adsl object>)`.
+#'        It's recommended to include `TealDataset` or `TealDatasetConnector` objects to
```
mhallal1, line 27:
```suggestion
#'        It is recommended to include `TealDataset` or `TealDatasetConnector` objects to
```

R/MAETealDataset.R
```diff
@@ -7,37 +7,33 @@
...
+#'        the `vars` list to preserve reproducibility. Please note that `vars`
+#'        are included to this object as local `vars` and they cannot be modified
+#'        within another dataset.
 #' @param metadata (named `list` or `NULL`) \cr
-#'   Field containing metadata about the dataset. Each element of the list
-#'   should be atomic and length one.
+#'        Field containing metadata about the dataset. Each element of the list
+#'        should be atomic and length one.
```
mhallal1, line 33:
```suggestion
#'        should be atomic and of length one.
```

R/MAETealDataset.R
mhallal1: same in `TealDataset`

DESCRIPTION
chlebowa: Dinakar was supposed to install the latest `roxygen2` but ran into some problems. It still works, though, tested in another package. And it can't be reversed, even if I downgrade my own `rozygen2`.

mhallal1 (approved):

PR #132, opened 2023-01-23
add cdisc information into JoinKeys

### LINKED ISSUE #120: Add cdisc_data information into join_keys
Related to https://github.com/insightsengineering/teal.data/issues/78 and may want to be done at the same time.

In order for tdata objects to be the universal data object in teal we need the cdisc information (i.e. parents) to be added into join_keys so that the information can be passed through to the filter panel. 

This issue is to:
1) Update join_keys class to allow parents to be included (should this be cdisc_join_keys)?
2) Update CDISCData$new/mutate to add the parents information into join_keys 

If once done this task it looks like we can remove CDISCTealData and just use TealData with cdisc_join_keys inside it then create an issue to do this and make changes downstream in teal and teal.slice.

mhallal1:
closes #120 

- removed `CDISCTealData` and replaced its functionality with `TealData`.
- removed `join_keys` mentions in `TealDataset` and `TealDatasetConnector`.
- updated wrappers `cdisc_data` and `teal_data` to set up join_keys by calling `JoinKeys` functionalities. They check if 
`join_keys` are passed otherwise they create them according to primary keys.
- deprecated `cdisc_data_file`
- Added tests to `test-joinKeys.R`
- Updated and added tests of `test-TealData`, `test-TealDataAbstract`, `test-cdisc_data` and `test-teal_data`.
- updated `teal.slice::init_filtered_data` to remove `CDISCTealData` S3 method and to account for `cdisc` status in the `TealData` S3 method: https://github.com/insightsengineering/teal.slice/pull/169


R/CDISCTealData.R
```diff
@@ -1,270 +0,0 @@
...
-#'                   ADTTE <- synthetic_cdisc_data('latest')$adtte\",
-#'           check = FALSE
-#'      )
-#'      # <code"
-#'   ),
-#'   con = file_example
-#' )
-#'
-#' cdisc_data_file(file_example)
-cdisc_data_file <- function(path, code = get_code(path)) {
```
nikolas-burkoff, line 266: should this be deprecated? happy to remove though

R/CDISCTealData.R
```diff
@@ -1,270 +0,0 @@
...
-  object <- object_file(path, "CDISCTealData")
```
nikolas-burkoff, line 267: object_file is not used anywhere else so can be removed as well?

tests/testthat/test-CDISCTealData.R
```diff
@@ -1,1205 +0,0 @@
-adsl_raw <- as.data.frame(as.list(setNames(nm = get_cdisc_keys("ADSL"))))
```
nikolas-burkoff, line 1: Are there any tests here which should have been TealData tests but were actually CDISCTealData ? If so we should move them rather than delete them

R/JoinKeys.R
```diff
@@ -164,12 +164,84 @@ JoinKeys <- R6::R6Class( # nolint
...
+      # check parent child relation
+      private$check_parent_child()
+
       invisible(self)
     }
   ),
   ## __Private Fields ====
   private = list(
     .keys = list(),
+    primary_keys = list(),
```
nikolas-burkoff, line 243: I don't think we need this as this can be stored in .keys as well (.keys[d, d]) and we can have a helper function `get_primary_keys(d)`

R/JoinKeys.R
```diff
@@ -331,13 +409,14 @@ mutate_join_keys.TealData <- function(x, dataset_1, dataset_2, val) { # nolint
...
 #' and `dataset_2`.
+#' @param parent (`character`) parent dataset
 #'
 #' @return object of class `JoinKeySet` to be passed into `join_keys` function.
 #'
 #' @seealso [join_keys()]
 #'
 #' @export
-join_key <- function(dataset_1, dataset_2, keys) {
+join_key <- function(dataset_1, dataset_2, keys, parent = character(0)) {
```
nikolas-burkoff, line 419: I don't think we should have parent here as it's not a function of the relationship between dataset_1 and dataset_2

R/teal_data.R
```diff
@@ -102,3 +99,34 @@ teal_data_file <- function(path, code = get_code(path)) {
...
+#' Add primary keys as join_keys to a dataset self
+#'
+#' @param data_objects (`list`) of `TealDataset`, `TealDatasetConnector` or `TealDataConnector` objects
+#' @param join_keys (`JoinKeys`) object
+#'
+get_primary_keys <- function(data_objects, join_keys) {
+  lapply(data_objects, function(obj) {
+    if (inherits(obj, "TealDataConnector")) {
+      sub_objs <- obj$get_items()
+      lapply(names(sub_objs), function(dataname) {
```
nikolas-burkoff, line 112: lets do this recursively:

```
get_primary_keys<- function(data_objects, join_keys) {
  lapply(data_objects, function(obj) {
    if (inherits(obj, "TealDataConnector") {
      get_primary_keys(obj$get_items(), join_keys)
    }
   else {
     if (length(...) {
       ...
    }
   }
  }

}

File: R/teal_data.R [hunk 7]
```diff
@@ -32,25 +32,22 @@
...
-
-  x <- if (any(is_cdisc)) {
-    CDISCTealData$new(..., check = check, join_keys = join_keys)
-  } else {
-    TealData$new(..., check = check, join_keys = join_keys)
+  if (inherits(join_keys, "JoinKeySet")) {
+    join_keys <- teal.data::join_keys(join_keys)
   }
 
+  get_primary_keys(data_objects, join_keys)
```
nikolas-burkoff, line 47: We should probably rename as get implies we are returning the primary keys rather than updating join_keys or we could just return the list of join_key() and do:
 x <- get_primary_keys(data_objects, join_keys) 
 lapply(x, function(y) join_keys$mutate(y))

R/cdisc_data.R
```diff
@@ -0,0 +1,127 @@
...
+    data_objects,
+    types = c("TealDataset", "TealDatasetConnector", "TealDataConnector")
+  )
+  if (inherits(join_keys, "JoinKeySet")) {
+    join_keys <- teal.data::join_keys(join_keys)
+  }
+
+  get_primary_keys(data_objects, join_keys)
+
+  if (length(join_keys$get_parents()) == 0) {
```
nikolas-burkoff, line 84: I think we should always do this not just if this condition is true as you may have only specified a few parents etc. 
It can error if the parents in the JoinKeys do not match the parents in the cdisc_dataset...

R/cdisc_data.R
```diff
@@ -0,0 +1,127 @@
...
+      tryCatch(
+        x$get_parent(),
+        error = function(cond) rep(character(0), length(x$get_datanames()))
+      )
+    }
+
+    new_parents <- unlist(
+      lapply(data_objects, function(x) {
+        if (inherits(x, "TealDataConnector")) {
+          lapply(x$get_items(), function(z) z$get_parent())
```
nikolas-burkoff, line 95: I think we can use recursion here see comment in teal_data...

R/cdisc_data.R
```diff
@@ -0,0 +1,127 @@
...
+        } else {
+          list(retrieve_parents(x))
+        }
+      }),
+      recursive = FALSE
+    )
+
+    names(new_parents) <- unlist(lapply(data_objects, function(x) {
```
nikolas-burkoff, line 103: Is there a way to create a named list without having to do it separately for `new_parents` and `parents`?

R/cdisc_data.R
```diff
@@ -0,0 +1,127 @@
+#' Data input for teal app
+#'
+#' @description `r lifecycle::badge("stable")`
+#' Function takes datasets and creates `CDISCTealData` object which can be used in `teal` applications.
```
nikolas-burkoff, line 4: Remove `CDISCTealData`

R/cdisc_data.R
```diff
@@ -0,0 +1,127 @@
...
+#' get_keys(test_data, "ADAE") # returns [1] "STUDYID" "USUBJID" "ASTDTM"  "AETERM"  "AESEQ"
+#' ```
+#' @inheritParams teal_data
+#' @param ... (`TealDataConnector`, `TealDataset` or
+#'   `TealDatasetConnector`) elements to include where `ADSL` data is mandatory.
+#' @param join_keys (`JoinKeys`) or a single (`JoinKeySet`)\cr
+#'   (optional) object with datasets column names used for joining.
+#'   If empty then it would be automatically derived basing on intersection of datasets primary keys
+#'
+#' @return a `CDISCTealData` object
```
nikolas-burkoff, line 27: Replace

nikolas-burkoff (commented):
Looking good a few comments to address then we'll take another look

R/CDISCTealData.R
mhallal1: It is used:
Screenshot from 2023-01-25 14-05-16

nikolas-burkoff: weird - wonder why I didn't find them...

R/cdisc_data.R
mhallal1: updated.

R/cdisc_data.R
mhallal1: removed.

R/cdisc_data.R
mhallal1: updated.

R/teal_data.R
mhallal1: updated.

R/JoinKeys.R
mhallal1: fair point, it is not used actually as we retrieve the `parents` from the `TealDataset` or `TealDatasetConnector` objects directly.

R/CDISCTealData.R
mhallal1: fair enough, deprecated with `deprecate_warn`.

R/cdisc_data.R
mhallal1: updated to always apply this and updated `joinKeys$set_parents()` to throw an error if the parents in the joinKeys do not match the parents in the cdisc_dataset.

tests/testthat/test-cdisc_data.R
```diff
@@ -0,0 +1,267 @@
...
+  file.remove(tmp_file)
+  testthat::expect_s3_class(tdf, "TealData")
+  testthat::expect_identical(
+    tdf$get_code(),
+    paste0(
+      "ADSL <- as.data.frame(as.list(setNames(nm = get_cdisc_keys(\"ADSL\"))))\n",
+      "adsl_raw <- as.data.frame(as.list(setNames(nm = get_cdisc_keys(\"ADSL\"))))\n",
+      "adsl <- cdisc_dataset(dataname = \"ADSL\", x = adsl_raw, ",
+      "code = \"ADSL <- as.data.frame(as.list(setNames(nm = get_cdisc_keys(\\\"ADSL\\\"))))\")\n",
+      "teal_data(adsl)"
```
nikolas-burkoff, line 241: I wonder if we should have cdisc_data(adsl) etc. (and in the teal_datafile case also have a test with it?)

tests/testthat/test-cdisc_data.R
```diff
@@ -0,0 +1,267 @@
...
+    )
+  )
+
+  result_to_compare <- do.call("cdisc_data", datasets)
+
+  test_relational_data_equal(result, result_to_compare)
+})
+
+testthat::test_that("cdisc_data_file loads the TealData object", {
+  tmp_file <- tempfile(fileext = ".R")
```
nikolas-burkoff, line 218: cdisc_data file tests need the lifecycle verbosity I think https://lifecycle.r-lib.org/reference/verbosity.html

tests/testthat/test-TealData.R
```diff
@@ -1,238 +1,342 @@
...
+  df1 <- dataset("df1", df1, keys = "id")
+  df2 <- dataset("df2", df2, keys = "df2_id")
 
-  expect_silent(
-    constructor_wrapper(
-      join_keys = join_keys(join_key("df_1", "df_2", c("x" = "u")))
-    )$check_metadata()
+  join_keys1 <- join_keys(join_key("df1", "df1", "id"), join_key("df2", "df2", "df2_id"))
+  data <- TealData$new(df1, df2, join_keys = join_keys1)
+  # primary keys are not taken from datasets when calling TealData$new(), these are only added if using wrappers
```
nikolas-burkoff, line 17: e.g. teal_data

tests/testthat/test-TealData.R
```diff
@@ -1,238 +1,342 @@
...
+  )
+})
 
-  expect_error(
-    constructor_wrapper(
-      join_keys = join_keys(join_key("df_1", "df_2", c("x" = "y", "v" = "v")))
-    )$check_metadata()
+testthat::test_that("TealData$print prints out expected output on basic input", {
+  adsl_raw <- as.data.frame(as.list(setNames(nm = get_cdisc_keys("ADSL"), object = list(1:3, letters[1:3]))))
+  adsl <- cdisc_dataset(
```
nikolas-burkoff, line 62: unless you are explicitly testing something cdisc related I'd use dataset in tests

tests/testthat/test-TealData.R
```diff
@@ -1,238 +1,342 @@
...
-testthat::test_that("Multiple connectors wrapped in cdisc_data", {
-  example_data_connector <- function(...) {
-    connectors <- list(...)
-    open_fun <- callable_function(library)
-    open_fun$set_args(list(package = "teal"))
-    con <- TealDataConnection$new(open_fun = open_fun)
-    TealDataConnector$new(connection = con, connectors = connectors)
-  }
+testthat::test_that("TealData$get_parents returns an empty list even when parents are specified", {
+  # parent specified - this information is passed on the wrappers level
```
nikolas-burkoff, line 218: I think comment should say:

Parent information is passed through the join_keys argument - this test does not create creating TealData with such a join_keys object set (it is only set for you if you call teal/cdisc_data)

tests/testthat/test-TealData.R
```diff
@@ -1,238 +1,342 @@
...
-    )
+
+  expect_silent(
+    constructor_wrapper(
+      join_keys = join_keys(join_key("df_2", "df_2", c("u" = "u")))
+    )$check_metadata()
+  )
+})
+
+testthat::test_that("TealData$check_metadata returns error when a column in th ekeys is not found", {
```
nikolas-burkoff, line 327:
```suggestion
testthat::test_that("TealData$check_metadata returns error when a column in the keys is not found", {
```

nikolas-burkoff (commented):
All looks good, a few minor comments and one in teal.slice then I suggest the team run some example apps (might be worth adding ones you've been using to the PR description if you haven't already) and then this can be merged

tests/testthat/test-TealData.R
mhallal1: done.

tests/testthat/test-TealData.R
mhallal1: done.

tests/testthat/test-TealData.R
mhallal1: done.

tests/testthat/test-TealData.R
mhallal1: done.

tests/testthat/test-cdisc_data.R
mhallal1: added.

tests/testthat/test-cdisc_data.R
mhallal1: updated here to use `cdisc_data`.
They are in fact the same tests as for `teal_data_file` but using `cdisc_data`.

tests/testthat/test-data_label.R
```diff
@@ -41,3 +41,15 @@ testthat::test_that("get_labels' column_labels is a named vector of the labels o
   attr(test[[2]], "label") <- test_labels[2]
   testthat::expect_equal(get_labels(test)$column_labels, stats::setNames(object = test_labels, nm = colnames(test)))
 })
+
+testthat::test_that("get_cdisc_keys returns column names present in the cached datasets", {
+  scda_data <- scda::synthetic_cdisc_data("latest")
+
+  testthat::expect_true(all(get_cdisc_keys("ADSL") %in% names(scda_data$adsl)))
```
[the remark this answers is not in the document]
mhallal1: done.

tests/testthat/test-TealDataAbstract.R
```diff
@@ -1,3 +1,481 @@
...
+  code = "ADSL <- as.data.frame(as.list(setNames(nm = get_cdisc_keys(\"ADSL\"))))"
+)
+adtte_cf <- callable_function(
+  function() {
+    as.data.frame(as.list(setNames(nm = get_cdisc_keys("ADTTE"))))
+  }
+)
+adtte <- dataset_connector("ADTTE", adtte_cf, keys = get_cdisc_keys("ADTTE"), vars = list(x = adsl))
+load_dataset(adsl)
+load_dataset(adtte)
```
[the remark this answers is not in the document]
mhallal1: done.

R/JoinKeys.R
```diff
@@ -164,12 +164,92 @@ JoinKeys <- R6::R6Class( # nolint
...
+              # both has the same parent -> parent keys
+              self$get(d1_parent, d1_parent)
+            } else {
+              # cant find connection - leave empty
+              next
+            }
+            self$mutate(d1, d2, fk)
+            duplicate_pairs <- append(duplicate_pairs, paste(d1, d2))
+          }
+        }
```
gogonzo, line 241: Since `mutate` changes keys to `dataset1` and `dataset2` in the same time, can we just make a simpler loop? If keys are taken from parents we don't need this `if` in the line 225, isn't it?

```suggestion
      datanames <- names(self$get())
      duplicate_pairs <- list()
      for (d1 in datanames) {
        d1_pk <- self$get(d1, d1)
        d1_parent <- self$get_parent(d1)
        if (length(d1_parent) > 0) {
          parent_pk <- self$get(d1_parent, d1_parent)
          self$mutate(d1, d1_parent, parent_pk)
        }
      }
      # check parent child relation
      private$check_parent_child()

      invisible(self)
```

R/cdisc_data.R
```diff
@@ -0,0 +1,165 @@
...
+  x <- TealData$new(..., check = check, join_keys = join_keys)
+
+  if (length(code) > 0 && !identical(code, "")) {
+    x$set_pull_code(code = code)
+  }
+
+  x$check_reproducibility()
+  x$check_metadata()
+  return(x)
+}
```
gogonzo, line 126: Unfortunately, this code overwrites manually specified keys.

```
library(scda)


latest_data <- synthetic_cdisc_data("latest")
ADSL <- latest_data$adsl
ADTTE <- latest_data$adtte
ADSL <- dplyr::rename(ADSL, SID = STUDYID) 
ADTTE <- dplyr::rename(ADTTE, SID = STUDYID)

# Example with keys
cdisc_data(
  cdisc_dataset("ADSL", ADSL,
    keys = c("STUDYID", "USUBJID")
  ),
  cdisc_dataset("ADTTE", ADTTE,
    keys = c("STUDYID", "USUBJID", "PARAMCD"),
    parent = "ADSL"
  ),
  join_keys = join_keys(
    join_key(
      "ADSL",
      "ADTTE",
      c("SID" = "SID", "USUBJID" = "USUBJID")
    )
  )
)
# Error in x$check_metadata() :
#  The join key specification requires dataset ADSL to contain the following columns: STUDYID, USUBJID
```

I wonder if setting default keys shoud be like this. Or maybe we should create `join_keys()` based on the datasets - this will give us default values of the join_keys in the beginning. Alternatively, you can make an `if` here to distinguish if the keys have been specified or not.

```suggestion
cdisc_data <- function(...,
                       join_keys = teal.data::join_keys(...),
                       code = "",
                       check = FALSE) {
  data_objects <- list(...)
  checkmate::assert_list(
    data_objects,
    types = c("TealDataset", "TealDatasetConnector", "TealDataConnector")
  )
  
  x <- TealData$new(..., check = check, join_keys = join_keys)

  if (length(code) > 0 && !identical(code, "")) {
    x$set_pull_code(code = code)
  }

  x$check_reproducibility()
  x$check_metadata()
  return(x)
}
```

gogonzo: Alternatively, cdisc_data could look like this:
```
cdisc_data <- function(...,
                       join_keys = teal.data::join_keys(),
                       code = "",
                       check = FALSE) {
  data_objects <- list(...)
  checkmate::assert_list(
    data_objects,
    types = c("TealDataset", "TealDatasetConnector", "TealDataConnector")
  )
  
  if (empty join keys) {
    join_keys <- initialize_datasets_keys(data_objects)
  }
  
  x <- TealData$new(..., check = check, join_keys = join_keys)
  if (length(code) > 0 && !identical(code, "")) {
    x$set_pull_code(code = code)
  }
  x$check_reproducibility()
  x$check_metadata()
  return(x)
}

```

R/JoinKeys.R
mhallal1: No, we need the double loop.
Imagine we have a `JoinKeys` object with 3 datasets inside, `ADSL`, `ADAE` and `ADTTE`. We define join_keys between `ADSL`  (the parent) and each of `ADAE` and `ADTTE` which turn out to be the same keys. In this case, the double loop creates join_keys between `ADAE` and `ADTTE` which are the same keys with the parent dataset. With the single loop, this won't be the case.

R/cdisc_data.R
mhallal1: The error returned by the app above makes sense and it accounted for because you are trying to define a `cdisc_dataset` with keys containing `STUDYID` where `STUDYID` is not in the data. Try changing `STUDYID` to `SID` and it will work smoothly.

NAMESPACE
```diff
@@ -41,7 +41,6 @@ S3method(mutate_data,TealDataAbstract)
 S3method(mutate_dataset,TealDataAbstract)
 S3method(mutate_dataset,TealDataset)
 S3method(mutate_dataset,TealDatasetConnector)
-S3method(mutate_join_keys,JoinKeys)
```
gogonzo, line 44: Why keys can't be mutated anymore? I suggest to keep it as it might be needed again in the near future. For example if we decide to have a function `get_default_cdisc_keys(c("ADSL", "ADTTE", "ADRS"))` which produces a `JoinKeys` object. We would like to be able to modify this object using `mutate_join_keys`.
For example
```r
get_default_cdisc_keys(c("ADSL", "ADTTE", "ADRS")) |>
  mutate_join_keys("ADSL", "ADSL", c("SID", "USID")) |>
  mutate_join_keys("ADTTE", "ADTTE", c("SID", "USID", "PARAMCD")) |>
  mutate_join_keys("ADSL", "ADTTE", c("SID", "USID"))
```

I think we might consider in the future something like this:
````r
get_default_cdisc_keys(c("ADSL", "ADTTE", "ADRS")) |>
  mutate_fk("ADTTE", "ADTTE", c("SID", "USID", "PARAMCD")) |>
  mutate_pk("ADSL", c("SID", "USID")) |>
  mutate_pk("ADTTE",  c("SID", "USID", "PARAMCD")) |>
  mutate_parent("ADTTE", "ADSL")
```

gogonzo (approved):
👍 Thanks for a huge simplification here

Issue #121, opened 2022-11-23
Output of TealData should be tdata


nikolas-burkoff:
Should be done with https://github.com/insightsengineering/teal/issues/783

We are aiming to make tdata the universal data object in the teal framework. This issue is to update the `teal.data` package so that the data coming out of it for `teal::init` is of type `tdata`

1) The class definition of tdata should be moved out of teal and into here
2) TealData$pull() should output a `tdata` object
3) For non DDL data a tdata object should be created from the datasets
4) tdata should have CodeClass

In the future #102 along with https://github.com/insightsengineering/teal/issues/764#issuecomment-1318403191 will simplify the teal.data api and push the use of tdata further into teal.data but that is not this issue

gogonzo:
Probably we need to store CodeClass in tdata as we need a way to subset code for selected datasets. `code` could be an expression only if we decide to not subset the code or if we figure out [this](https://github.com/insightsengineering/teal.data/issues/37)

Issue #97, opened 2022-08-25
review check() method 


donyunardi:
There are evidence of check method logic is buggy:

- https://github.com/insightsengineering/teal.data/issues/17
- https://github.com/insightsengineering/teal.data/issues/13

Need to find the root cause, review design, or refine action items.

asbates:
I'd like to suggest considering a name change. `check` is not at all obvious. I think it should be something more like `check_code` or `check_dataset_code`.

PR #138, opened 2023-02-24
Tdata in tealdata@main

mhallal1:
closes #https://github.com/insightsengineering/teal.data/issues/121
depends on https://github.com/insightsengineering/teal.slice/pull/212 and https://github.com/insightsengineering/teal/pull/808

-Moved `tdata` from `teal` to `teal.data`
-Added `get_tdata` method to `TealData`
-Updated downstreams to use `TealData$get_tdata()` output instead of a `TealData` object
-Updated tests

TODO:
-figure the `code` issue


R/tdata.R
```diff
@@ -0,0 +1,188 @@
...
+#' isolate(data[["iris"]]())
+#'
+#' # Get code
+#' isolate(get_code_tdata(data))
+#'
+#' # Get metadata
+#' get_metadata(data, "iris")
+#'
+#' @export
+new_tdata <- function(data, code = "", join_keys = NULL, metadata = NULL, check = FALSE, label) {
```
mhallal1, line 48: the `check` argument here is not in its perfect place and is only passed downstream inside `FilteredData` to be called in `teal::get_datasets_code` to evaluate if an warning message should be returned in the SRC or not.
An alternative to this would be not to pass `check` at all but rather to evaluate if `check` is TRUE or FALSE upstream and throw a warning message in the console. This will not return a warnings message in the SRC as it is done now but will avoid us the passing an unnecessary argument all through the teal app setup.
@nikolas-burkoff @donyunardi @lcd2yyz

nikolas-burkoff: https://github.com/insightsengineering/teal.data/issues/97

asbates:
NEWS update? But I guess not if this is internal only.

R/tdata.R
```diff
@@ -0,0 +1,188 @@
...
+
+
+#' Wrapper for `get_code.tdata`
+#' This wrapper is to be used by downstream packages to extract the code of a `tdata` object
+#'
+#' @param data (`tdata`) object
+#'
+#' @return (`character`) code used in the `tdata` object.
+#' @export
+get_code_tdata <- function(data) {
```
asbates, line 136: I don't understand why there are 2 functions here. I could maybe see it if `get_code.tdata` was not exported, but it is. As a user I would be confused as to which one I should use.

R/TealData.R
```diff
@@ -193,6 +193,28 @@ TealData <- R6::R6Class( # nolint
       private$join_keys$get_parents()
     },
 
+    #' @description
+    #' returns the `tdata` object.
+    #'
+    #' @return (`tdata`) object of the datasets.
+    get_tdata = function() {
```
asbates, line 200: I don't understand the need for this method. And I don't see it used anywhere throughout Teal.

donyunardi:
I want to confirm the reason we have `get_tdata` and `new_tdata`.

1.  `get_tdata` is being used in `teal::srv_teal` function to pass tdata create FilteredData object.
    ```r
    datasets <- teal.slice::init_filtered_data(raw_data()$get_tdata())
    ```
    
1.  In `init_filtered_data.tdata` generic class, we unpack the `tdata` so we can feed it in to `init_filtered_data.default`. At this point, the datasets in step 1 will return a `FilteredData` object.

1.  then `new_tdata` is being used in `teal::srv_nested_tabs` when`.datasets_to_data` function is run to create `tdata` from `FilteredData` object.
    ```r
    data <- .datasets_to_data(modules, datasets, trigger_data)
    ```

If this is true, why do we need to feed `tdata` during `init_filtered_data` during step 1?
It felt like we're not doing anything special to the object during step 2.

nikolas-burkoff:
> If this is true, why do we need to feed tdata during init_filtered_data during step 1? It felt like we're not doing anything special to the object during step 2.

So the idea here is eventually we will not have a `TealData` class at all but only `tdata` (and a lightweight class to generate `tdata` in a delayed way) - it's way too big a change to do that in one go - and will involve big changes to `teal::init`'s API. This PR is a first step along that path - being able to create a `tdata` from `TealData` directly

Eventually the data flow in `teal::init` will go something like:

1) User specify `tdata` object (or instructions as to how to do it in a delayed way)
2) Filter panel is created from this `tdata` object
3) Filter panel has a `datasets$get_tdata()` method (so we don't need `.datasets_to_data` method) which returns a `tdata` object with the filter applied
4) Modules take `tdata` object

Part 4) is already on main. Part 2) is in this collection of PRs, part 3) is waiting for `teal.slice` changes (and tbh `.datasets_to_data` is effectively doing this). And part 1) is very much at the early stages here

After all these changes, in order to use/test/maintain `teal::init`, individual modules or the filter panel you only need to create one `tdata` object which is used to create all these classes. 

This should simplify things a lot - a year ago in order to test a module you needed to create a `FilteredData` object from which you needed to have created a `[CDISC]TealData` object which meant there was strong coupling between all the components and it was harder to understand and maintain.  

-----

Another benefit we can introduce here is that at the moment pulling the data can only occur once per shiny session - we can change this as some apps want to be able to pull the data multiple times - and also this can then remove the splash screen for DDL and the additional reactivity cycle coming from insertUI - see [this](https://github.com/insightsengineering/teal/tree/534_nested_tabs%40main) old branch for inspiration 

-----

There are some complexities at this point which need to be handled - hence trying to do just this baby step first: 

1) the `check` argument to `cdisc_data`  is currently passed all the way through so that the modules "show r code" in the can give a message saying "the app developer did not check the data" reproducing this behaviour is painful as it complicates our `tdata` object needing an extra `check` argument for imho very little benefit

Do we actually want to keep this behaviour (esp given the open "look at `check` again issue") - or can we avoid doing this - this is a question for @lcd2yyz 

2) The show r code for each module includes the code needed to get the data but not all of the data, only the data used in that module. This is good if we are treating each module's SRC as independent and we want to run the r code outside of the app (if there's 20 datasets but this module only uses 2 then no need to get hold of the other 18...) but adds a lot of complexity (as we need to know which lines of preprocessing code are used for which datasets) - and again would make tdata much more complex. If the SRC is mainly meant as a record of what could be run and we can live with all dataset preprocessing included in the SRC then the implementation becomes massively simpler. 

Happy to talk about the above points in a meeting to explain in more detail. The correct solution to these really is teal 1.0:

Instead of everything (check, preprocessing code, filtering code, filter_panel_api for the formatted filter states) being passed into the modules to be included in the report/show r code you have the modules passing their code snippet into teal::init to handle the SRC/report cards. In practice this means SRC becomes teal getting information from teal.data, the filter panel and the modules and combining them together. 

Teal is the right place for these things to live: "why does a module need to know the code used to filter the data that is being passed into it" and "why does every module need filter_panel_api passed into it purely to get the filter information in the report card - when the module itself has no concept of filtered data - it just has "data"?" 

This then would allow much better reports (with say a "Data processing" section containing the getting the data and then "Outputs" section containing the module outputs - and the R code for each output would not need another copy of the "getting data" section as it currently does.

but I think this is quite a long way away - a decisions could be made to pause this work and attempt work on teal 1.0 first, or to abandon completely etc.

R/TealData.R
nikolas-burkoff: So this is used in teal 
image

R/tdata.R
asbates: What about axing `get_code.tdata` and just use `get_code_tdata` then? It feels like an unnecessary complexity to have both.

If we keep `get_code.tdata` though, the generic documentation should be updated. Right now the generic reads `@param x ([`TealDatasetConnector`] or [`TealDataset`]). If of class `character` will be treated as file to read.` It should also say `tdata` in this case.

R/TealData.R
asbates: Ah OK, missed that.

lcd2yyz:
> There are some complexities at this point which need to be handled - hence trying to do just this baby step first:
> 
> 1. the `check` argument to `cdisc_data`  is currently passed all the way through so that the modules "show r code" can give a message saying "the app developer did not check the data" reproducing this behaviour is painful as it complicates our `tdata` object needing an extra `check` argument for imho very little benefit
> 
> Do we actually want to keep this behaviour (esp given the open "look at `check` again issue") - or can we avoid doing this - this is a question for @lcd2yyz

I don't know the history of this message, I can only guess that it was added to limit potential liability on app developer. 
I believe once we implement https://github.com/insightsengineering/teal/issues/778, we should no longer need to worry about passing this message. So I'm fine with removing this `check`, consistent with the decision to remove this https://github.com/insightsengineering/teal.widgets/issues/48


> Teal is the right place for these things to live: "why does a module need to know the code used to filter the data that is being passed into it" and "why does every module need filter_panel_api passed into it purely to get the filter information in the report card - when the module itself has no concept of filtered data - it just has "data"?"
> 
> This then would allow much better reports (with say a "Data processing" section containing the getting the data and then "Outputs" section containing the module outputs - and the R code for each output would not need another copy of the "getting data" section as it currently does.

Agree with this, but sounds like major refactor again, something we need to come back to in the future. Can you help open a new issue to log down all your thoughts about this?

nikolas-burkoff:
> Agree with this, but sounds like major refactor again, something we need to come back to in the future. Can you help open a new issue to log down all your thoughts about this?

So in some sense all the refactor work we've been doing for teal has been leading towards this - my thoughts are in agreement with @gogonzo and he has our thoughts together with a diagram [here](https://github.com/insightsengineering/teal/milestone/5) (the list of issues in the milestone is not complete)

R/tdata.R
nikolas-burkoff: > What about axing get_code.tdata and just use get_code_tdata then? It feels like an unnecessary complexity to have both.

Fair enough

gogonzo:
Closing, we are choosing different solution

PR #161, opened 2023-08-15
Refactor

gogonzo:
Part of https://github.com/insightsengineering/teal.data/issues/102
Other PRs https://github.com/insightsengineering/teal/pull/899 https://github.com/insightsengineering/teal.slice/pull/440

## install
<details>
<summary>Installation code </summary>

```r
staged.dependencies::dependency_table(
  project = "insightsengineering/teal@https://github.com",
  project_type = "repo@host",
  ref = "refactor",
  verbose = 1
) |> staged.dependencies::install_deps()
```
</details>

## ddl class

<details><summary>unfold</summary>

Small class which contains `ui` and `server`. `server` calls `run_ddl` which evaluates a `code` using input and create `teal_data` object. Function `ddl` accepts several arguments which are used to generate proper `ui` and `server`.

tdata-DDL process (3)

Above diagram depict how class works. Please follow steps below and look at the diagram. 

1. One needs to specify `ddl` arguments to fit someone's needs. Take for example `custom_connector`. Let's assume that we need to call `scda::synthetic_cdisc_data('latest')` for two datasets.

```r
data_code = paste(
    "adsl <- scda::synthetic_cdisc_data('latest')$adsl",
    "adtte<- scda::synthetic_cdisc_data('version')$adtte",
    collapse = "\n"
  )
)
```

Imagine that above datasets are on the remote server so we need to call `open_conn` function and close after pull. Imagine that `open_conn` and `close_conn` are in some external package. 

```r
code = paste(c(
  "open_conn(username = input$username, password = input$password)",
  "adsl <- scda::synthetic_cdisc_data('latest')$adsl",
  "adtte<- scda::synthetic_cdisc_data('version')$adtte",
  "close_conn()"
  ),
  collapse = "\n"
)
```

`username` and `password` should be specified by the app user. To do so we need to create `ui` which will create `username` and `password` inputs which are used during evaluation of above code.

```r
ui = function(id) {
  ns <- NS(id)
  tagList(
    textInput(ns("username"), label = "Username"),
      passwordInput(ns("password"), label = "Password"),
      actionButton(ns("submit"), label = "Submit")
  )
}
```

If we have `ui` we need also `server` to utilize input. When submit button is clicked all `input` values are passed to `ddl_run` where all magic happens. `ddl_run` executes the code and substitutes all `input$<name>` with equivalent `input_list` element. 

```r
submit_button_server <- function(id) {
  moduleServer(id, function(input, output, session) {
    tdata <- eventReactive(input$submit, {
[...]

PR #167, opened 2023-09-18
ddl alternative

chlebowa:
`ddl` implementation alternative to #161 .
Complemented by [this PR](https://github.com/insightsengineering/teal/pull/922).

In order to simplify the user (app dev) experience, I tried to streamline the logic.

In order to create a `ddl` connector module, one has to:
1. use `input_template` to create the module: enumerate input widgets
2. provide a function, `on_submit`, to be run when `"submit"` button is clicked; function takes input values wrapped in a list called `input` and body refers to input values with `input$<value>` or `input[["<value>"]]`
3. optionally provide mask for input values that will be used in code of resulting `tdata` object
4. specify names of data sets for compatibility with `teal` (I don't like it)
5. optionally specify join keys as one would previously, for compatibility with `teal`; defaults to empty `teal.data::join_keys()`

When inputs are submitted, `on_submit` is passed to a function that extracts the body, substitutes `input` placeholders with input values and evaluates the code to obtain data sets in a separate environment. Then it replaces the input values in the code with ones provided in `mask` (if any) and uses the environment and the masked code to create `tdata`.

Much like in the solution proposed on branch `refactor`, the user provides code to obtain data sets and replacements for input values, and data is created in separate environment, which is then used to create `tdata` with masked code.

Unlike that solution, the user specifies everything in one place, rather than having to define module ui, module server that runs a post-processing function, the post-processing function itself, etc. This is easier to understand **for me**.
Another difference is that the user provides code as code with `input$` references, not text with `glue` syntax (`{ value }`). This is done move focus to masking rather than have the user think about "online" and "offline" arguments. It also uses pure base R.

#### MOCK DATABASE CONNECTION
```
pullme <- function(username, password) {
  if (username == "user" && password == "pass") {
    message("connection established")
  } else {
    stop("invalid credentials")
  }
}
closeme <- function() {
  message("connection closed")
}
```
#### MODULE DEFINITION
```
library(shiny)

thefun <- function(input) {
  on.exit(try(closeme()))
  pullme(username = input$user, password = input$pass)
  adsl <- scda::synthetic_cdisc_data('latest')$adsl
  adtte <- scda::synthetic_cdisc_data('latest')$adtte
}
themask <- list(
  user = quote(askpass("who are you?")),
  pass = quote(askpass("password please"))
)
module <- input_template(
  on_submit = thefun,
  mask = themask,
  datanames = c("adsl", "adtte"),
  textInput("user", "username", value = "user", placeholder = "who goes there?"),
  passwordInput("pass", "password", value = "pass", placeholder = "friend or foe?"),
  actionButton("submit", "get it")
)
```
#### AN APP
```
devtools::load_all("../teal.slice")
devtools::load_all("../teal")
devtools::load_all(".")

ui <- fluidPage(
  tagList(
    module$ui("id"),
    uiOutput("val")
  )
)
server <- function(input, output, session) {
  tdata <- module$server("id")
  output[["value"]] <- renderPrint({
    tdata()
  })
  output[["code"]] <- renderPrint({
    cat(teal.code::get_code(tdata()), sep = "\n")
  })
  output[["val"]] <- renderUI({
    req(tdata())
    tagList(
      verbatimTextOutput("value"),
      verbatimTextOutput("code")
    )
  })
}
if (interactive()) shinyApp(ui, server)
```

#### A TEAL APP
```
funny_module <- function (label = "Filter states", datanames = "all") {
  checkmate::assert_string(label)
  module(
    label = label,
    datanames = datanames,
    ui = function(id, ...) {
      ns <- NS(id)
      div(
        h2("The following filter calls are generated:"),
        verbatimTextOutput(ns("filter_states")),
        verbatimTextOutput(ns("filter_calls")),
        actionButton(ns("reset"), "reset_to_default")
      )
    },
    server = function(input, output, session, data, filter_panel_api) {
      checkmate::assert_class(data, "tdata")
      observeEvent(input$reset, set_filter_state(filter_panel_api, default_filters))
      output$filter_states <-  renderPrint({
        logger::log_trace("rendering text1")
        filter_panel_api %>% get_filter_state()
      })
      output$filter_calls <- renderText({
        logger::log_trace("rendering text2")
        attr(data, "code")()
      })
    }
  )
}

devtools::load_all("../teal.slice")
devtools::load_all("../teal")
devtools::load_all(".")

app <- init(
  data = module,
  modules = modules(
    funny_module("funny1"),
    funny_module("funny2", datanames = "adtte") # will limit datanames to ADTTE and ADSL (parent)
  )
)
shinyApp(app$ui, app$server)
```

### DIFF
--- a/R/__ddl_by_AC.R
+++ b/R/__ddl_by_AC.R
@@ -191,7 +191,6 @@ with_substitution <- function(fun, mask, join_keys) {
 
 
 library(shiny)
-devtools::load_all()
 
 # mock database connection
 pullme <- function(username, password) {
@@ -220,10 +219,17 @@ themask <- list(
 module <- input_template(
   on_submit = thefun,
   mask = themask,
+  datanames = c("adsl", "adtte"),
   textInput("user", "username", value = "user", placeholder = "who goes there?"),
   passwordInput("pass", "password", value = "pass", placeholder = "friend or foe?"),
   actionButton("submit", "get it")
 )
+
+
+devtools::load_all("../teal.slice")
+devtools::load_all("../teal")
+devtools::load_all(".")
+
 ui <- fluidPage(
   tagList(
     module$ui("id"),
@@ -247,3 +253,43 @@ server <- function(input, output, session) {
   })
 }
 if (interactive()) shinyApp(ui, server)
+
+
+
+# funny_module <- function (label = "Filter states", datanames = "all") {
+#   checkmate::assert_string(label)
+#   module(
+#     label = label,
+#     datanames = datanames,
+#     ui = function(id, ...) {
+#       ns <- NS(id)
+#       div(
+#         h2("The following filter calls are generated:"),
+#         verbatimTextOutput(ns("filter_states")),
+#         verbatimTextOutput(ns("filter_calls")),
+#         actionButton(ns("reset"), "reset_to_default")
+#       )
+#     },
+#     server = function(input, output, session, data, filter_panel_api) {
+#       checkmate::assert_class(data, "tdata")
+#       observeEvent(input$reset, set_filter_state(filter_panel_api, default_filters))
+#       output$filter_states <-  renderPrint({
+#         logger::log_trace("rendering text1")
+#         filter_panel_api %>% get_filter_state()
+#       })
+#       output$filter_calls <- renderText({
+#         logger::log_trace("rendering text2")
+#         attr(data, "code")()
+#       })
+#     }
+#   )
+# }
+#
+# app <- init(
+#   data = module,
+#   modules = modules(
+#     funny_module("funny1"),
+#     funny_module("funny2", datanames = "adtte") # will limit datanames to ADTTE and ADSL (parent)
+#   )
+# )
+# shinyApp(app$ui, app$server)
--- a/R/__ddl_by_AC.R
+++ b/R/__ddl_by_AC.R
@@ -26,6 +26,8 @@
 #' @param ... any number of `shiny.tag`s
 #' @param on_submit function to run after clicking the `submit` button, see `Details`
 #' @param mask optional list specifying how to mask the code run by `on_submit`, see `Details`
+#' @param datanames character vector of names of data sets created; required for compatibility with `teal` apps
+#' @param join_keys `join_keys` object specifying relationships between data sets; defaults to `teal.data::join_keys()`
 #' @return A`reactive` expression returning a `tdata` object.
 #'
 #' #' @examples
@@ -134,6 +136,7 @@ input_template <- function(..., on_submit, mask, datanames, join_keys) {
 #'
 #' @param fun a function that takes exactly one argument, `input`, which is a named list
 #' @param mask optional named list to specify code masking; see `input_template` for details
+#' @param join_keys optional `join_keys` object; see `input_template` for details
 #'
 #' @return
 #' A `tdata` object containing variables that were created in the body of `fun`
@@ -230,66 +233,66 @@ devtools::load_all("../teal.slice")
 devtools::load_all("../teal")
 devtools::load_all(".")
 
-ui <- fluidPage(
-  tagList(
-    module$ui("id"),
-    uiOutput("val")
-  )
-)
-server <- function(input, output, session) {
-  tdata <- module$server("id")
-  output[["value"]] <- renderPrint({
-    tdata()
-  })
-  output[["code"]] <- renderPrint({
-    cat(teal.code::get_code(tdata()), sep = "\n")
-  })
-  output[["val"]] <- renderUI({
-    req(tdata())
-    tagList(
-      verbatimTextOutput("value"),
-      verbatimTextOutput("code")
-    )
-  })
-}
-if (interactive()) shinyApp(ui, server)
-
-
-
-# funny_module <- function (label = "Filter states", datanames = "all") {
-#   checkmate::assert_string(label)
-#   module(
-#     label = label,
-#     datanames = datanames,
-#     ui = function(id, ...) {
-#       ns <- NS(id)
-#       div(
-#         h2("The following filter calls are generated:"),
-#         verbatimTextOutput(ns("filter_states")),
-#         verbatimTextOutput(ns("filter_calls")),
-#         actionButton(ns("reset"), "reset_to_default")
-#       )
-#     },
-#     server = function(input, output, session, data, filter_panel_api) {
-#       checkmate::assert_class(data, "tdata")
-#       observeEvent(input$reset, set_filter_state(filter_panel_api, default_filters))
-#       output$filter_states <-  renderPrint({
-#         logger::log_trace("rendering text1")
-#         filter_panel_api %>% get_filter_state()
-#       })
-#       output$filter_calls <- renderText({
-#         logger::log_trace("rendering text2")
-#         attr(data, "code")()
-#       })
-#     }
-#   )
-# }
-#
-# app <- init(
-#   data = module,
-#   modules = modules(
-#     funny_module("funny1"),
-#     funny_module("funny2", datanames = "adtte") # will limit datanames to ADTTE and ADSL (parent)
+# ui <- fluidPage(
+#   tagList(
+#     module$ui("id"),
+#     uiOutput("val")
 #   )
 # )
-# shinyApp(app$ui, app$server)
+# server <- function(input, output, session) {
+#   tdata <- module$server("id")
+#   output[["value"]] <- renderPrint({
+#     tdata()
+#   })
+#   output[["code"]] <- renderPrint({
+#     cat(teal.code::get_code(tdata()), sep = "\n")
+#   })
+#   output[["val"]] <- renderUI({
+#     req(tdata())
+#     tagList(
+#       verbatimTextOutput("value"),
+#       verbatimTextOutput("code")
+#     )
+#   })
+# }
+# if (interactive()) shinyApp(ui, server)
+
+
+
+funny_module <- function (label = "Filter states", datanames = "all") {
+  checkmate::assert_string(label)
+  module(
+    label = label,
+    datanames = datanames,
+    ui = function(id, ...) {
+      ns <- NS(id)
+      div(
+        h2("The following filter calls are generated:"),
+        verbatimTextOutput(ns("filter_states")),
+        verbatimTextOutput(ns("filter_calls")),
+        actionButton(ns("reset"), "reset_to_default")
+      )
+    },
+    server = function(input, output, session, data, filter_panel_api) {
+      checkmate::assert_class(data, "tdata")
+      observeEvent(input$reset, set_filter_state(filter_panel_api, default_filters))
+      output$filter_states <-  renderPrint({
+        logger::log_trace("rendering text1")
+        filter_panel_api %>% get_filter_state()
+      })
+      output$filter_calls <- renderText({
+        logger::log_trace("rendering text2")
+        attr(data, "code")()
+      })
+    }
+  )
+}
+
+app <- init(
+  data = module,
+  modules = modules(
+    funny_module("funny1"),
+    funny_module("funny2", datanames = "adtte") # will limit datanames to ADTTE and ADSL (parent)
+  )
+)
+shinyApp(app$ui, app$server)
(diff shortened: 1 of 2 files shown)


R/__ddl_by_AC.R
```diff
@@ -0,0 +1,230 @@
+#' Create input module.
+#'
+#' Template for creating a simple module that will put up input widgets and produce `tdata`.
+#'
+#' Define any inputs necessary to connect to a remote data source and produce data,
+#' as well as a function (`on_submit`) that will create the desired data sets.
+#'
+#' One of the inputs must be an action button (or action link) called "submit".
```
m7pr, line 8:
```suggestion
#' One of the inputs must be an action button (or action link) called `"submit"`.
```

R/__ddl_by_AC.R
```diff
@@ -0,0 +1,230 @@
...
+#' When clicked, the `on_submit` function will be run.
```
m7pr, line 9:
```suggestion
#' When clicked, `on_submit` function will run.
```

R/__ddl_by_AC.R
```diff
@@ -0,0 +1,230 @@
...
+#'
+#' `on_submit` must take one argument called `inputs`,
+#' which will be a list of all input elements defined in the UI function except `"submit"`.
```
m7pr, line 12:
```suggestion
#' `on_submit` takes one argument called `inputs`,
#' which is a list of all input elements defined in the UI function, except `"submit"`.
```

R/__ddl_by_AC.R
```diff
@@ -0,0 +1,230 @@
...
+#' The function body must contain all code necessary to obtain the desired data sets and nothing else.
```
m7pr, line 13:
```suggestion
#' `on_submit` body should contain the code necessary to obtain the desired datasets and nothing else.
```

R/__ddl_by_AC.R
```diff
@@ -0,0 +1,230 @@
...
+#' Do not include a return statement, simply assign your data sets to appropriate variables.
```
m7pr, line 14:
```suggestion
#' Do not return any value, just assign your datasets to appropriate variables. See examples.
```

R/__ddl_by_AC.R
```diff
@@ -0,0 +1,230 @@
...
+#'
+#' Clicking "submit" will run the function provided in `on_submit`.
```
m7pr, line 16:
```suggestion
#' Clicking `"submit"` will run the function provided in `on_submit`.
```

R/__ddl_by_AC.R
```diff
@@ -0,0 +1,230 @@
...
+#' The obtained data sets will be packed into a `tdata` object.
+#' The body of `on_submit` will be recorded in the resulting `tdata`.
+#'
+#' The `mask` argument can be used to mask input values used as arguments in the recorded code.
```
m7pr, line 20:
```suggestion
#' The `mask` argument can be used to mask input values used as arguments in the recorded code. So that sensitive parts of code (like password or username) are substituted with `mask`ed values.
```

R/__ddl_by_AC.R
```diff
@@ -0,0 +1,230 @@
...
+#'
+#' Input elements will be put in a div of class `connector-input`.
```
m7pr, line 22:
```suggestion
#' Input elements will be put in a `connector-input` div.
```

R/__ddl_by_AC.R
```diff
@@ -0,0 +1,230 @@
...
+#'
+#' @param ... any number of `shiny.tag`s
+#' @param on_submit function to run after clicking the "submit" button, see `Details`
```
m7pr, line 25:
```suggestion
#' @param on_submit a function to run after clicking the `"submit"` button, see `Details`
```

R/__ddl_by_AC.R
```diff
@@ -0,0 +1,230 @@
...
+#' @param mask optional list specifying how to mask the code run by `on_submit`, see `Details`
```
m7pr, line 26:
```suggestion
#' @param mask an optional list specifying how to mask the code run by `on_submit`, see `Details`
```

R/__ddl_by_AC.R
```diff
@@ -0,0 +1,230 @@
...
+#' @return A`reactive` expression returning a `tdata` object.
```
m7pr, line 27:
```suggestion
#' @return A `reactive` expression returning a `tdata` object.
```

R/__ddl_by_AC.R
```diff
@@ -0,0 +1,230 @@
...
+  checkmate::assert_true(
+    is.element("submit", inputIds),
+    .var.name = "A \"submit\" element is specified."
+  )
+
+  submit <- unlist(eval(args[[which(inputIds == "submit")]]))
+  submit_class <- submit[grep("class$", names(submit))]
+  checkmate::assert_true(
+    grepl("action-button", submit_class),
+    .var.name = "The \"submit\" element has class \"action-button\"."
```
m7pr, line 76:
```suggestion
    .var.name = "The \"submit\" element class: \"action-button\"."
```

R/__ddl_by_AC.R
```diff
@@ -0,0 +1,230 @@
...
+  )
+}
+
+
+# cc <- list(
+#   quote(pullme("user", "pass", "versicolor")),
+#   quote(closeme())
+# )
+# arglist <- quote(list(password = askpass("big pas")))
+# mask_code(cc, list(pullme = arglist))
```
m7pr, line 120:
```suggestion

```

chlebowa: oops :o

R/__ddl_by_AC.R
```diff
@@ -0,0 +1,230 @@
...
+
+#' wrap a function so that it returns its code in addition to the result
```
m7pr, line 115:
```suggestion
#' Extend function's result with body returned as an additional attribute
```

R/__ddl_by_AC.R
```diff
@@ -0,0 +1,230 @@
...
+#' @param fun a function
+#' @return A function that works just like `fun` but adds its body to the result as the `code_used` attribute.
```
m7pr, line 117:
```suggestion
#' @return An input `fun` function, but with its body added to the `code_used` attribute.
```

m7pr:
Hey @chlebowa I can't give any feedback to the code, as it's extremely condensed, clean and contains a lot of advanced tricks that reduces the complexity of the implementation. As always, a solid through work!

When it comes to the interface and user experience, I prefer this approach than #161 as I do see benefits of having everything in one place (`input_template`). This is very clear and short in terms of user specification and usage. It also gives a very direct and simple interface to actions happening on `on_submit` action and objects getting `mask`ed in the final output. And we don't need to deal with `glue` (yuk!). 

You have my axe!


A small question, I wonder if there is a way to specify `input_template` so that if pulls needed data for the app, after you click `Accept` button. And before you do, you don't see any data being used in teal app. This way we could solve https://github.com/insightsengineering/teal/issues/812

```r
module <- input_template(
  h5('Below is a consens statement, that you need to accept before using the app. Do you accept?')
  actionButton("submit", "Accept"),
  on_submit = function(input) {
    # before on_submit data is NULL
    data <- pull(input$data) # ??
  }
)
```

chlebowa:
I just pushed a major overhaul of roxygen docs, which should have been there from the beginning but slipped my mind during the multiple rewrites this file has seen ober the last two weeks. Sorry for the inconvenience and thank you @m7pr for bringing it to my attention.

R/__ddl_by_AC.R
```diff
@@ -0,0 +1,237 @@
...
+      } else if (is.name(fun_body)) {
+        fun_body
+      } else {
+        stop("with_substitution: don't know ho to handle this kind of function body")
+      }
+
+    # Convert calls to strings and substitute argument references by bquote references.
+    code_strings <- vapply(code, deparse1, character(1L))
+    code_strings <- gsub("(input\\$)(\\w+)", "\\.(\\2\\)", code_strings)
+    code_strings <- gsub("(input\\[\\[\")(\\w+)(\"\\]\\])", "\\.(\\2\\)", code_strings)
```
m7pr, line 168: @chlebowa I would make sure those regexes work on all platforms. Looks like system specific?

chlebowa: As far as I know this is just R-specific, it looks a little weird because every slash has to be escaped. `\w` and backtracing are widely recognized.

m7pr: coolio

chlebowa:
> A small question: is there a way to use `input_template` as a welcome-popup that prevents data beoing loaded into the app before you accept a message appearing over a submit buton? So that we could solve this [insightsengineering/teal#812](https://github.com/insightsengineering/teal/issues/812)

I think that issue is about a popup on the whole app, regardless of whether the data is remote or local, so `ddl` would not reliably solve it.

R/__ddl_by_AC.R
averissimo: There may be some edge cases where users use `purrr::plucK(input, "something")` or something else.

chlebowa: True, `get, for instance. I'm not sure we have to worry about it, considering people mosty use `$`.

Now that I think about it, some NSE tricks would probably go unnoticed :thinking:

chlebowa:
Update: this version of `ddl` can now be used in a `teal` app :tada: 
Unfortunately, some new arguments had to be added to `input_template` but there is nothing I can do about it for now.

I wonder what to do with errors raised by the user-provided code. Currently they are intercepted and passed to `validate` calls but they can still be lost before the user has a chance to see them. Compare the normal app to the `teal` one. 
A solution would be to run the code in a `qenv` (or `tdata`) to take advantage of the `qenv.error` class to propagate errors to app outputs.

Requires `teal` branch `alternative@refactor@main`. `staged.dependencies` should handle it but one never knows.

R/__ddl_by_AC.R
```diff
@@ -0,0 +1,298 @@
...
+    # Use bquote to obtain code with input values and masking values.
+    code_input <- lapply(code_strings, function(x) do.call(bquote, list(str2lang(x), call_args)))
+    code_masked <- lapply(code_strings, function(x) do.call(bquote, list(str2lang(x), mask)))
+
+    # Evaluate input code in separate environment.
+    env <- new.env()
+    eval(as.expression(code_input), env)
+    # Create `tdata` with masked code.
+    new_tdata(as.list(env), code = as.expression(code_masked), keys = join_keys)
```
gogonzo, line 191: We need a way to change `code_masked` when creating `tdata` object

```suggestion
    new_tdata(as.list(env), code = as.expression(code_masked), keys = join_keys)
```


There is a case where in the interactive session one can use `login()` function which opens a dialog, while for shiny session authentication is handled by the module and no `login()` function is used. It means that initial `code_masked` looks like this

```r
data <- connection_package::read_data(...)
```

and returned should look like:

```r
connection_package::login()
data <- connection_package::read_data(...)
```

R/__ddl_by_AC.R
```diff
@@ -0,0 +1,298 @@
...
+    moduleServer(id, function(input, output, session) {
+      result <- eventReactive(input[["submit"]], {
+        inputs <- sapply(setdiff(inputIds, "submit"), function(x) input[[x]], simplify = FALSE)
+        tryCatch(
+          do.call(tracked_request, list(inputs)),
+          error = function(e) validate(need(FALSE, sprintf("Error: %s", e$message)))
+        )
+      })
+      result
+    })
```
gogonzo, line 114: We need a possibility to specify this `server`. For example to control hierarchical inputs.
```suggestion
  server <- function(id) {
    moduleServer(id, function(input, output, session) {
      result <- eventReactive(input[["submit"]], {
        inputs <- sapply(setdiff(inputIds, "submit"), function(x) input[[x]], simplify = FALSE)
        tryCatch(
          do.call(tracked_request, list(inputs)),
          error = function(e) validate(need(FALSE, sprintf("Error: %s", e$message)))
        )
      })
      result
    })
```

gogonzo:
Having just `input_template` without option to specify server limits possibilities. Besides, I like argument name `on_submit` but I'm indifferent whether it's a call, character or a function. The biggest issue I have with this approach is that call which creates a `tdata` is not open for modifications (like postprocess_fun in my PR) - this limits potential use cases.

PR #172, opened 2023-10-05
Fix the unit tests for the teal_data refactor

vedhav:
Fixes the broken tests due to the teal_data refactor


gogonzo (approved):

gogonzo:
Thanks @vedhav

Issue #169, opened 2023-09-29
change `new_teal_data` arguments


gogonzo:
**Acceptance criteria:**
Constructor method should have variants for a signature `data = "list"` and `data = "environment"`.

Constructor should work with following call:

```r
new_teal_data(
  data = list(iris2 = iris2),
  code = quote(iris2 <- iris) 
)

new_teal_data(
  data = as.environment(list(iris2 = iris2)),
  code = quote(iris2 <- iris) 
)
```

gogonzo:
Closing this PR. `new_teal_data` will be a private constructor. We will expose `teal_data/cdisc_data` only.

PR #178, opened 2023-10-18
introduce teal_data class

gogonzo:
Part of:
- https://github.com/insightsengineering/teal/issues/937

Closes:
- https://github.com/insightsengineering/teal.data/issues/169

Other PRs:
- https://github.com/insightsengineering/teal/pull/925 
- https://github.com/insightsengineering/teal.slice/pull/470
- https://github.com/insightsengineering/teal.gallery/pull/75 

## install

```r
staged.dependencies::dependency_table(
  project = "insightsengineering/teal@https://github.com",
  project_type = "repo@host",
  ref = "teal_data@main",
  verbose = 1
) |> staged.dependencies::install_deps()
```

## teal_data class

tdata-qenv, tdata inheritance

- `teal_data` class (teal_data-class.R) is en extension of `qenv`. `teal_data` contains extra slots `datanames` and `join_keys`. Besides, all `teal.code` methods works with `teal_data` object (eval_code, print etc.)
- Old `TealData$server` returns `teal_data` now - to have single object type in teal (not in `teal_module`)
- `cdisc_data()` and `teal_data()` return `TealData` or `teal_data`

```r
ADSL <- synthetic_cdisc_data("latest")$adsl
ADTTE <- synthetic_cdisc_data("latest")$adtte
ADRS <- synthetic_cdisc_data("latest")$adrs

data <- cdisc_data(
  ADSL = ADSL,
  ADTTE = ADTTE,
  ADRS = ADRS,
  code = quote({
    ADSL <- synthetic_cdisc_data("latest")$adsl
    ADTTE <- synthetic_cdisc_data("latest")$adtte
    ADRS <- synthetic_cdisc_data("latest")$adrs
  })  
)

data |> eval_code(quote(a <- ADSL))
```


R/cdisc_data.R
```diff
@@ -68,58 +68,71 @@ cdisc_data <- function(...,
...
+    }
 
-  new_parents_fun <- function(data_objects) {
-    lapply(data_objects, function(x) {
+    new_parents <- unlist(new_parents_fun(data_objects), recursive = FALSE)
+
+    names(new_parents) <- unlist(lapply(data_objects, function(x) {
       if (inherits(x, "TealDataConnector")) {
-        unlist(new_parents_fun(x$get_items()), recursive = FALSE)
+        lapply(x$get_items(), function(z) z$get_dataname())
```
pawelru, line 109: To signal that it's double (not: triple) nested loop
```suggestion
        lapply(x$get_items(), function(y) y$get_dataname())
```

R/cdisc_data.R
```diff
@@ -68,58 +68,71 @@ cdisc_data <- function(...,
...
 
-  if (length(code) > 0 && !identical(code, "")) {
-    x$set_pull_code(code = code)
+    x$check_reproducibility()
+    x$check_metadata()
+    x
+  } else {
+    if (!checkmate::test_names(names(data_objects), type = "named")) {
+      stop("Dot (`...`) arguments on `teal_data()` must be named.")
+    }
```
pawelru, line 133: Do we want to support ["dynamic dots"](https://rlang.r-lib.org/reference/dyn-dots.html)?

I am thinking here about re-collecting dots with `rlang:: dots_list` which

R/get_join_keys.R
```diff
@@ -0,0 +1,20 @@
+#' Function to get join keys from a `` object
+#' @param data `` - object to extract the join keys
+#' @return Either `JoinKeys` object or `NULL` if no join keys
+#' @export
+get_join_keys <- function(data) {
```
pawelru, line 5: If you introduce a new method please check for possible replacement of old method for consistency reasons
```
❯ grep -r "get_join_keys()" --include="*.R" .
./R/module_teal_with_splash.R:        keys = data$get_join_keys()
./R/module_teal_with_splash.R:            keys = data$get_join_keys()
./R/module_nested_tabs.R:    datasets$get_join_keys(),
```

R/teal.data.R
```diff
@@ -13,6 +13,6 @@
 #' @import shiny
 #' @importFrom digest digest
 #' @importFrom stats setNames
-#' @importFrom shinyjs show
+#' @importFrom shinyjs useShinyjs
```
pawelru, line 16: why this?

R/teal_data-class.R
```diff
@@ -0,0 +1,103 @@
+setOldClass("JoinKeys")
+
+#' Reproducible data class.
+#'
+#' Reproducible data class basing on [`teal.code::qenv-class`].
+#' Don't interact with slots directly, it is suggested to use methods instead
+#' (see `methods(class = "teal_data")`).
```
pawelru, line 7: This feels a little aggressive. How about: please see methods to interact with class' slots.

R/teal_data-class.R
```diff
@@ -0,0 +1,103 @@
...
+#'
+#' @slot code (`character`) representing code necessary to reproduce the environment.
+#'  To extract the `code` please use [get_code()].
+#' @slot env (`environment`) environment which content was generated by the evaluation
+#'  of the `code` slot. To extract variables from the environment please use [get_var()] or [`[[`].
+#' @slot id (`integer`) random identifier of the code element to make sure uniqueness
+#'  when joining.
+#' @slot warnings (`character`) the warnings output when evaluating the code.
+#' To extract the `warnings` use [get_warnings()].
+#' @slot messages (`character`) the messages output when evaluating the code
```
pawelru, line 20: can you inherit roxygen params description from parent class to avoid repetition?

R/teal_data-class.R
```diff
@@ -0,0 +1,103 @@
...
+})
+
+#' @rdname new_teal_data
+#' @export
+setMethod(
+  "new_teal_data",
+  signature = c(data = "list", code = "character", keys = "ANY"),
+  function(data, code, keys = join_keys(), datanames = names(data)) {
+    new_env <- rlang::env_clone(list2env(data), parent = parent.env(.GlobalEnv))
+    lockEnvironment(new_env, bindings = TRUE)
```
pawelru, line 67: why it's needed? maybe worth to put it in the in-line comment?

R/to_relational_data.R
```diff
@@ -120,9 +120,3 @@ to_relational_data.list <- function(data) {
 to_relational_data.MultiAssayExperiment <- function(data) { # nolint
   teal_data(dataset("MAE", data))
 }
-
-#' @keywords internal
-#' @export
-to_relational_data.TealData <- function(data) { # nolint
-  data
-}
```
pawelru, line 128: why?

This potentially could be quite dangerous - you would have to track down and analyse all the usage of `to_relational_data()`. From what I recall its used in apply through a list of elements of unknown class. If it happens that the element is `TealData` then after this change we will have an error.

DESCRIPTION
```diff
@@ -33,6 +33,7 @@ Imports:
     shinyjs,
     stats,
     teal.logger (>= 0.1.1),
+    teal.code (>= 0.4.1.9003),
```
pawelru, line 36: can you please also add it below in `Config/Needs/verdepcheck` for verdepcheck tests

pawelru:
This is quite fundamental change how the app is being created. Please update validate and update examples to reflect the new way. This potentially has to be done in other repos as well.

R/to_relational_data.R
gogonzo: This function is used in `teal::init` I removed above method and decided to make [if condition instead](https://github.com/insightsengineering/teal/blob/e8c9d8623840edb71fbb38e6ed31052f4d7c961b/R/init.R#L118)

If I didn't make this if, I would have to introduce `to_relational.teal_data` also. `to_relational` will be removed after internal release together with all `TealData` related stuff

R/teal_data-class.R
gogonzo: environment is locked to prevent making changes when somebody tries to modify `obj@env` directly instead of using `eval_code` or `within`. Don't know though if lock really works. I'll investigate this

R/cdisc_data.R
pawelru: Actually I think that this could be a very neat feature implemented at relatively low cost.

R/teal_data-class.R
gogonzo: There is no `@inheritSlots` roxygen tag unfortunatelly

R/teal.data.R
gogonzo: this is why https://github.com/daattali/shinyjs/blob/67b269df8f07860c6be8dd10fb90575eee840c8d/NAMESPACE#L31

`shinyjs` exports method show which mask generic method! But don't worry, `shinyjs` dependency will be removed soon after internal release.

R/teal_data-class.R
pawelru: Have you tried `@inheritSection slots` (or rather `Slots`)?

R/cdisc_data.R
gogonzo: Yes, but at this moment this list could be named (new approach) and unnamed (old approach). So I can't just assert directly when obtaining `...`

```
  data_objects <- list(...)
  if (checkmate::test_list(data_objects, types = c("TealDataConnector", "TealDataset", "TealDatasetConnector")) {
     # make TealData
   } else {
     if (!checkmate::test_names(names(data_objects), type = "named")) {
      stop("Dot (`...`) arguments on `teal_data()` must be named.")
    }
    # make teal_data
   
   }

```

pawelru: You would have to collect dots twice
```
if (old approach condition) {
  objs <- dots_list(..., .named = FALSE, <other>)
  ...
} else {
  objs <- dots_list(..., .named = TRUE, <other>)
  ...
}
```

gogonzo: Three times actually, because logical condition is based on output from `...`

R/teal_data-class.R
gogonzo: Doesn't seem to work properly. Look, added slots doesn't merge with inherited section.
 
Skärmavbild 2023-10-20 kl  10 39 21

R/cdisc_data.R
pawelru: ok how about this:
```
objs <- dots_list(..., .named = FALSE, <other>)
if (condition(objs)) {
  ...
} else {
  objs <- dots_list(..., .named = TRUE, <other>)
  ...
}
```

Then after we hard deprecate old approach it would be just this:
```
  objs <- dots_list(..., .named = TRUE, <other>)
  ...
```

R/teal_data-class.R
pawelru: Indeed. It looks like it's a missing feature in roxygen. Let's leave it then. Thanks for checking

DESCRIPTION
```diff
@@ -57,7 +58,7 @@ Config/Needs/verdepcheck: rstudio/shiny, mllg/checkmate,
     yaml=vubiostat/r-yaml, rstudio/bslib, tidyverse/dplyr, yihui/knitr,
     tidyverse/magrittr, bioc::MultiAssayExperiment, tidyverse/readr,
     rstudio/reticulate, rstudio/rmarkdown, bioc::SummarizedExperiment,
-    r-lib/testthat, r-lib/withr
+    r-lib/testthat, r-lib/withr, insightsengineering/teal.code
```
pawelru, line 61: To keep the alphabetical order from Imports

```suggestion
    insightsengineering/teal.code, r-lib/testthat, r-lib/withr
```

R/teal_data-class.R
```diff
@@ -0,0 +1,74 @@
...
+#' @rdname new_teal_data
+#' @keywords internal
+new_teal_data <- function(data, code = character(0), keys = join_keys(), datanames = names(data)) {
+  checkmate::assert_list(data)
+  checkmate::assert_class(keys, "JoinKeys")
+  checkmate::assert_character(datanames)
+  if (!any(is.language(code), is.character(code))) {
+    stop("`code` must be a character or language object.")
+  }
+  code <- teal.code:::format_expression(code)
```
pawelru, line 59: you have just moved it there

```suggestion
  code <- format_expression(code)
```

R/teal_data-class.R
gogonzo: I've checked lockEnvironment. Works as expected and covered by tests here

https://github.com/insightsengineering/teal.data/blob/bb6e0ffc350b9939adb306ae8cdd71229a21b015/tests/testthat/test-teal_data.R#L86

chlebowa:
Even though this branch is green, these tests fail on my machine :thinking: 
```
Failure (test-TealDatasetConnector.R:567:3): code_dataset_connector - library calls
`data <- cdisc_data(adsl, adtte, adrs, check = TRUE)` did not produce any warnings.
```
```
Failure (test-TealDatasetConnector.R:567:3): code_dataset_connector - library calls
`data <- cdisc_data(adsl, adtte, adrs, check = TRUE)` did not produce any warnings.
```

tests/testthat/test-to_relational_data.R
```diff
@@ -105,9 +105,9 @@ test_that("to_relational_data accepts a function returning a TealDataset as inpu
 test_that("to_relational_data accepts a MultiAssayExperiment as input", {
   utils::data(miniACC, package = "MultiAssayExperiment")
   mae <- miniACC
-  output_dataset <- to_relational_data(mae)
+  output_dataset <- to_relational_data_wrapper(mae)
```
m7pr, line 108: @gogonzo why do we need `to_relational_data_wrapper` if it's just a function calling `to_relational_data`
```r
to_relational_data_wrapper <- function(data) {
  to_relational_data(data)
}
```
When it comes to the name - not really a wrapper, but rather an identity function. Wrapper would do some slight preparations or modifications.

gogonzo: This will be removed soon. Let's skip the discussion about this

m7pr: coolio

R/get_join_keys.R
```diff
@@ -0,0 +1,53 @@
...
+`get_join_keys<-` <- function(data, dataset_1, dataset_2 = NULL, value) {
+  UseMethod("get_join_keys<-", data)
+}
+
+
+#' @rdname get_join_keys
+#' @inheritParams mutate_join_keys
+#' @export
+`get_join_keys<-.JoinKeys` <- function(data, dataset_1, dataset_2 = NULL, value) {
+  data
```
averissimo, line 46: Consider adding comment to give extra context on why this assignment method does nothing.

```suggestion
  # The reason this passthrough method is defined is to prevent a warning message
  # The assignment is performed by `get_join_keys.JoinKeys` and `[<-.JoinKeys` combination
  #  as well as `JoinKeys` being an R6 class
  data
```

R/get_join_keys.R
```diff
@@ -0,0 +1,53 @@
...
+}
+
+#' @rdname get_join_keys
+#' @export
+`get_join_keys<-.teal_data` <- function(data, dataset_1, dataset_2 = NULL, value) {
+  data
```
averissimo, line 52: Consider adding comment to give extra context on why this assignment method does nothing.

```suggestion
  # The reason this passthrough method is defined is to prevent a warning message
  # The assignment is performed by `get_join_keys.teal_data` and `[<-.JoinKeys` combination
  #  as well as `JoinKeys` being an R6 class
  data
```

tests/testthat/test-TealDatasetConnector.R
```diff
@@ -564,7 +564,7 @@ testthat::test_that("code_dataset_connector - library calls", {
     label = "ADRS dataset"
   )
 
-  data <- cdisc_data(adsl, adtte, adrs, check = TRUE)
+  expect_warning(data <- cdisc_data(adsl, adtte, adrs, check = TRUE), "deprecated")
```
averissimo, line 567: Just as @chlebowa said, this test is only successful once every 6 hours.

Using `lifecycle::expect_deprecated` works best to detect deprecated functions

```suggestion
  lifecycle::expect_deprecated(data <- cdisc_data(adsl, adtte, adrs, check = TRUE))
```

averissimo (commented):
`teal_data` looks very good at the current state!

Some minor comments, below.. The `test/testhat/test-TealDatasetConnector.R` is the only one that seems important to merge.

---

My only big grip is one layer down with using S4 classes, in part by having exposed slots in `qenv` and by consequence on `teal_data` class.

It seems more beneficial to keep those fields as private and using active bindings to "expose" those as read-only.

However, converting `qenv` to `R6` is a whole new big project and may introduce caveats as well _(hence the use of S4... this predates my joining of the project and I'm not familiar with all the reasons)_.

R/teal_data-class.R
```diff
@@ -0,0 +1,78 @@
...
+#'
+#' @name teal_data-class
+#' @rdname teal_data-class
+#'
+#' @slot code (`character`) representing code necessary to reproduce the environment.
+#'  To extract the `code` please use [get_code()].
+#' @slot env (`environment`) environment which content was generated by the evaluation
+#'  of the `code` slot. To extract variables from the environment please use [get_var()] or [`[[`].
+#' @slot id (`integer`) random identifier of the code element to make sure uniqueness
+#'  when joining.
```
chlebowa, line 17:
```suggestion
#' @slot id (`integer`) random identifier assigned to each line of code in `@code`. Used internally.
```

R/teal_data-class.R
```diff
@@ -0,0 +1,78 @@
...
+#' @slot warnings (`character`) the warnings output when evaluating the code.
+#' To extract the `warnings` use [get_warnings()].
```
chlebowa, line 19:
```suggestion
#' @slot warnings (`character`) warnings raised when evaluating code. Access with [`get_warnings`].
```

R/teal_data-class.R
```diff
@@ -0,0 +1,78 @@
...
+#' @slot messages (`character`) the messages output when evaluating the code
```
chlebowa, line 20:
```suggestion
#' @slot messages (`character`) messages raised when evaluating code.
```

R/teal_data-class.R
```diff
@@ -0,0 +1,78 @@
...
+#' @slot join_keys (`JoinKeys`) object.
+#'  To extract the `join_keys` use [get_join_keys()].
```
chlebowa, line 22:
```suggestion
#' @slot join_keys (`JoinKeys`) object specifying joining keys for data sets in `@env`.
#'                 Access or modify with [`get_join_keys`].
```

R/teal_data-class.R
chlebowa, line 1: `teal_data` is instantiated with an empty string in `@code`, I believe that should be `character(0)`.
```
> td <- teal_data()
> teal.code::get_code(td)
[1] ""

Also, `teal.data::get_code` raises an error. I believe a method is missing. May be solved when `get_code` is removed from `teal.data` and masking no longer occurs.
```
> teal.data::get_code(td)
Error in get_code.default(td) : 
  Assertion on 'x' failed: Must be of type 'character', not 'teal_data'.
```

File: R/teal_data-class.R (hunk 5 above)
Comment (chlebowa) [reply R/teal_data-class.R:7]: > I let you suggest the whole phrase.

Don't mind if I do :wink:

File: R/teal_data-class.R [hunk 20]
```diff
@@ -0,0 +1,78 @@
...
+#' @slot datanames (`character`) names of datasets in `env`. Needed when non-dataset
+#'  objects are needed in the `env` slot.
+#'  To extract the `datanames` use [get_dataname()].
```
chlebowa, line 25:
```suggestion
#' @slot datanames (`character`) vector of names of data sets in `@env`.
#'                 Used internally to distinguish them from auxiliary variables.
#'                 Access or modify with [`datanames`].
```

R/teal_data-class.R
```diff
@@ -0,0 +1,78 @@
...
+#'   Needed when non-dataset objects are needed in the `env` slot.
+#' @rdname new_teal_data
+#' @keywords internal
+new_teal_data <- function(data,
+                          code = character(0),
+                          join_keys = join_keys(),
+                          datanames = union(names(data), names(join_keys$get()))) {
+  checkmate::assert_list(data)
+  checkmate::assert_class(join_keys, "JoinKeys")
+  if (is.null(datanames)) datanames <- character(0) # todo: allow to specify
```
chlebowa, line 58: When is it `NULL`?

R/utils.R
```diff
@@ -150,3 +150,38 @@ get_teal_bs_theme <- function() {
     bs_theme
   }
 }
+
+#' Format expression to string
+#' Convert any expression to a single character vector
+#' @param code (`language`, `expression`, `character`)
+#' @return `character(1)`
+#' @keywords internal
+format_expression <- function(code) {
```
chlebowa, line 159: Why redefine function in this package rather than use `teal.code:::format_expression`?

R/teal_data.R
```diff
@@ -3,8 +3,12 @@
...
 #' Universal function to pass data to teal application
 #'
-#' @param ... (`TealDataConnector`, `TealDataset`, `TealDatasetConnector`)\cr
-#'   objects
+#' @param ... (`TealDataConnector`, `TealDataset`, `TealDatasetConnector`, `any`)\cr
+#'  - When one of the `Teal*` objects are provided, then function returns `TealData` object.
+#'    This way of specifying data is deprecated and will be removed in the next release.
+#'  - From version 0.4.0, one can provide any object as a named argument and function will
+#'    return `teal_data` object. Objects provided in `...` will be stored in `teal_data` environment
+#'    under the same name as the argument name.
```
chlebowa, line 11: This should only contain accepted argument types.

R/teal_data.R
```diff
@@ -13,52 +17,63 @@
 #'  code included in the object definitions actually produces those objects.
 #'  If `check` is true and preprocessing code is empty an error will be thrown.
 #'
-#' @return (`TealData`)
+#' @return (`TealData` or `teal_data`) object
```
chlebowa, line 20: Specify when.

R/teal_data.R
```diff
@@ -13,52 +17,63 @@
...
+#' teal_data(
+#'   x1 = iris,
+#'   x2 = mtcars,
+#'   code = quote({
+#'     x1 <- iris
+#'     x2 <- mtcars
+#'   })
 #' )
-#'
-#' teal_data(x1, x2)
```
chlebowa, line 33: #' # specify data only
#' teal_data(x1, x2)
#' 
#' # specify data and code
#' teal_data(
#'   x1 = iris,
#'   x2 = mtcars,
#'   code = quote({
#'     x1 <- iris
#'     x2 <- mtcars
#'   })
#' )
#' 
#' # evaluate code in empty object
#' teal_data() |> within(x1 <- iris) |> within(x2 <- mtcars)

R/teal.data.R
```diff
@@ -11,8 +11,9 @@
 
 # Fix R CMD check notes
 #' @import shiny
+#' @importFrom rlang %||%
 #' @importFrom digest digest
 #' @importFrom stats setNames
-#' @importFrom shinyjs show
+#' @importFrom shinyjs useShinyjs
```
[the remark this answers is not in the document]
gogonzo: All declared imports should be used. `useShinyjs` is used inside of R6 so compiler doesn't recognise this. BTW I've made this change only because `shinyjs show` override `utils` generic for `show` method

NEWS.md
```diff
@@ -1,5 +1,13 @@
 # teal.data 0.3.0.9007
 
+### Enhancements
+* Simplified `join_key` to better support primary keys.
```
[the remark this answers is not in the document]
gogonzo: It is the `join_key` which have been improved

R/JoinKeys.R
```diff
@@ -370,6 +373,78 @@ join_keys <- function(...) {
...
+
+    if (name %in% names(default_cdisc_keys)) {
+      # Set default primary keys
+      keys_list <- default_cdisc_keys[[name]]
+      join_keys[name] <- keys_list$primary
+
+      if (!is.null(keys_list$parent) && !is.null(keys_list$foreign)) {
+        join_keys[name, keys_list$parent] <- keys_list$foreign
+      }
+    }
```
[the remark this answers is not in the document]
gogonzo: Not true. If `name` is not null then previous `if-else` doesn't exit (don't return NULL)

R/teal_data-class.R
gogonzo: When initializing empty. Then, names of the join_keys  and names of the data are empty (NULL)

tests/testthat/test-TealDatasetConnector.R
chlebowa: Oooh, that's what it was!

R/teal.data.R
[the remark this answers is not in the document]
chlebowa: This is why we have `.rectify_dependencies_check` in `teal.slice` [here](https://github.com/insightsengineering/teal.slice/blob/686f1f76ac6130325df9314fea419800c139f35a/R/zzz.R#L37).

Overrides? Not masks? Prefixing doesn't work? :exploding_head:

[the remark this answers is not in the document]
chlebowa: Can't wait :laughing:

R/teal_data.R
gogonzo: There are no restrictions on accepted argument types. Technically it is limited by teal.slice only, but it shouldn't be a concern of this package

R/JoinKeys.R
```diff
@@ -441,11 +524,15 @@ mutate_join_keys.TealData <- function(x, dataset_1, dataset_2, val) { # nolint
 #' @seealso [join_keys()]
 #'
 #' @export
-join_key <- function(dataset_1, dataset_2, keys) {
+join_key <- function(dataset_1, dataset_2 = NULL, keys) {
```
chlebowa, line 527: Remove default `NULL`?

gogonzo: or even `dataset_2 = dataset_1`

R/teal_data.R
chlebowa: I mean this is not the place to explain different return values. Sorry or not being clear.

gogonzo: Please propose something which makes sense at this moment. Feel free to make a commit 👍

R/teal_data-class.R
```diff
@@ -0,0 +1,88 @@
...
+#' These slots should never be accessed directly, use the provided get/set functions.
+#'
+#' As code is evaluated in `teal_data`, messages and warnings are stored in their respective slots.
+#' If errors are raised, a `qenv.error` object is returned.
+#'
+#' @name teal_data-class
+#' @rdname teal_data-class
+#'
+#' @slot env (`environment`) environment containing data sets and possibly auxiliary variables
+#'  Access variables with [get_var()] or [`[[`].
```
chlebowa, line 18:
```suggestion
#'  Access variables with [`get_var`] or [`[[`].
```

R/teal_data-class.R
```diff
@@ -0,0 +1,88 @@
...
+#' @slot code (`character`) representing code necessary to reproduce the contents of `@env`.
+#'  Access with [`get_code`].
+#'  No setter provided. Evaluate code to append code to the slot.
+#' @slot id (`integer`) random identifier assigned to each line of code in `@code`. Used internally.
+#' @slot warnings (`character`) the warnings output when evaluating the code.
+#' @slot warnings (`character`) warnings raised when evaluating code.
+#'  Access with [get_warnings()].
+#' @slot messages (`character`) messages raised when evaluating code.
+#' @slot join_keys (`JoinKeys`) object specifying joining keys for data sets in `@env`.
+#'  Access or modify with [get_join_keys()].
```
chlebowa, line 29:
```suggestion
#'  Access or modify with [`get_join_keys`].
```

R/teal_data-class.R
```diff
@@ -0,0 +1,88 @@
...
+#' @slot datanames (`character`) vector of names of data sets in `@env`.
+#'  Used internally to distinguish them from auxiliary variables.
+#'  Access or modify with [datanames()].
```
chlebowa, line 32:
```suggestion
#'  Access or modify with [`datanames`].
```

gogonzo: I see, `datanames` is a topic

R/teal_data-class.R
gogonzo: what is wrong with `[get_var()]`

chlebowa: It's a function call, usually docs mention function _names_, and in `code` format.

R/teal.data.R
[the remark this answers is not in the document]
m7pr: > Who cares? In two weeks it won't be discussed anymore ;]

Best argument ever : P

R/teal_data-class.R
m7pr: +1 for using () in function names

chlebowa: The distinction between topic and function seems unclear to me. Is a function NOT a topic? What is a topic then? 

Here's my view:
The docs should read "Change datanames slot with datanames function."
One changes the slot with `datanames<object> <- <new_datanames>` not with `datanames()`. If you insist on the latter, you lose consistency with `[[`.

Look at help pages for `environment`, `get`, `assign`, etc. Hardly a `foo()` in sight.

chlebowa: + I always thought that table is an example, not a directive.

chlebowa: Ok, so a topic would be a help page and one can be dedicated to a concept or a function or a group of functions or anything in between. Correct?

R/teal_data-class.R
gogonzo: `teal_code` initializes with empty character now. `get_code` masking will have to be solved soon when removing old classes/method.

chlebowa: Good enough for me :+1:

chlebowa (approved):
Let's go!

PR #181, opened 2023-10-25
add `datanames` function

chlebowa:
Adds function `datanames` to get or set the contents of the `@datanames` slot in `teal_data` objects.


gogonzo (approved):
👍

PR #183, opened 2023-10-30
Deprecation of `get_code.default`

### LINKED ISSUE #173: Deprecate get_code.default
This `get_code` method is a very old function which can read the preprocessing (everything before teal::init) to obtain code needed to reproduce the data. Function is used for example in [longitudinal example app](https://github.com/insightsengineering/teal.gallery/blob/6e3e42420aec2090bb469b5403c33830632c35d8/longitudinal/app.R#L422) to avoid writing code to `cdisc_data(code)` argument. This function has a very limited functionality and albo collides with `get_code` method for `qenv` and `teal_data`. We should have only one `get_code` method in the whole framework.

Function should be deprecated now, and removed with old TealData related functions after next release.

m7pr:
Close #173

### DIFF
--- a/R/get_code.R
+++ b/R/get_code.R
@@ -112,6 +112,8 @@ get_code.default <- function(x,
   checkmate::assert_flag(exclude_comments)
   checkmate::assert_flag(read_sources)
 
+  lifecycle::deprecate_warn("0.3.1", "get_code.default()")
+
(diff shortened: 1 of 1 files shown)


gogonzo:
We will hard deprecate old functions and move them to one file after ddl is merged. Please close the PR

PR #186, opened 2023-11-09
move import to package level file and update pre-commit

donyunardi:
* Moving `@import` to package level file for consistency.
* Update pre-commit setup.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,7 +25,7 @@ repos:
           - shiny
           - shinyjs
           - stats
-          - insightsengineering/teal.logger
+          - teal.logger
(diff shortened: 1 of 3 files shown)


chlebowa:
I prefer keeping `@import` tags next to the items that necessitate them myself. Then again, this is a case where the class definition is inherited so there is no `teal_data` without `teal.code`. But then again again, if that were to change, this `@import` tag will probably be overlooked and remain in the package docs as an orphan.

donyunardi:
Okay, fair enough.
Let's keep it closer to the class itself.

Looks like pre-commit additional dependencies also updated during the refactor. 

Closing PR

PR #197, opened 2023-11-22
fix get_code_dependency for empty input

m7pr:
Close https://github.com/insightsengineering/teal/issues/972

### DIFF
--- a/R/utils-code_dependency.R
+++ b/R/utils-code_dependency.R
@@ -313,6 +313,10 @@ get_code_dependency <- function(code, names) {
   checkmate::assert_multi_class(code, classes = c("character", "expression"))
   checkmate::assert_character(names)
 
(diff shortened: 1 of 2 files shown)


R/utils-code_dependency.R
```diff
@@ -313,6 +313,10 @@ get_code_dependency <- function(code, names) {
   checkmate::assert_multi_class(code, classes = c("character", "expression"))
   checkmate::assert_character(names)
 
+  if (identical(code, character(0)) || identical(code, "")) {
```
averissimo, line 316: Small suggestion to treat string that only has whitespace as an empty string

```suggestion
  if (identical(code, character(0)) || identical(trimws(code), "")) {
```

averissimo (approved):
Works well with the set of examples that were used to detect this issue :+1:

PR #199, opened 2023-11-22
trimws to code at "" if

m7pr:
Have had an automerge on https://github.com/insightsengineering/teal.data/pull/197 and could not apply suggestions


chlebowa (approved):
I'm not sure this is needed but certainly won't hurt.

gogonzo (approved):
👍

PR #201, opened 2023-11-23
192 Improve `get_code_dependency` (Code Parser)

### LINKED ISSUE #192: Potential simplification of `get_code_dependency`
### Feature description

A follow-up after https://github.com/insightsengineering/teal.code/pull/146 related to those 2 comments
https://github.com/insightsengineering/teal.code/pull/146#discussion_r1347383335
https://github.com/insightsengineering/teal.code/pull/146#discussion_r1347231031

In the current format, `get_code_dependency()` works on `code_dependency()` that creates a structure of the dependency of objects found within the code. `code_dependency()` returns a named list of length 3. There is an idea that this list can be simplified to length 1. This list contains

- `occurence` named list (names after all found symbols in the code) with integer vectors specifying in which call a symbol appeared
- `cooccurrence` a list of length equal to the number of calls in input code, each element contains names of symbols appearing in this call. symbols are ordered so that the first element is the affected object
- `effects` a named list (the same as `occurence`) specifying where `linksto` tags appeared for a specific object

1. `occurence` and `effects` could be merged into the same list
2. `occurence` and `effects` could be merged into `cooccurence` - if there is just one object in an element of the list, this is just an occurence or an effect. if there are more then this is `cooccurrence`. we need to sort out what if there is a `cooccurrence` that has an effect which will actually impact 2 objects (the affected and the one tagged with `linksto`
3. what if `linktso` has multiple objects ?
4. `detect_symbol` can be then used once (for `cooccurrence` and `occurrence`)

`cooccurence` could be a list of length equal to the number of call, where each element is a list containing 2 fields
- influenced - objects being impacted by this line
- influencers - object having an impact on this line

m7pr:
**Closes #192**

# Overview 

The **Code Parser** feature accepts code in text form and outputs all necessary code to recreate any object from the original input. 

If the code contains side effects that don't precisely specify the influenced object(s), users can add a comment with the tag `# @linksto object_name1 object_name2` to explicitly identify the affected objects. This ensures that the side effects are included when generating the code needed to create the objects. **This is also why the input needs to be a `character`, not `expression` or `language` because comments are preserved in `character` and are removed in `expression` and `language`.**

# Historical Background

- The `CodeDepends` package was evaluated to check if it provided the features required to accomplish our goals ([#238](https://github.com/insightsengineering/teal/issues/238)).
    - However, `CodeDepends` was found to have excessive dependencies and was deemed more than what we needed.
    - Consequently, we developed the current solution, named **Code Parser**`
        - It is based on the graph dependency concept in `CodeDepends` and utilizes `utils::getParseData`.
        - `utils::getParseData` works on a `srcref` data.frame structure, 
        - `srcref` is created by calling `attr(parse(text = code, keep.source = TRUE), 'srcref')`
        - That's why  `code` is limited to character or `expression` with `srcref` attribute.
- The **Code Parser** was initially introduced in `teal.code` ([PR #146](https://github.com/insightsengineering/teal.code/pull/146)), designed to take the `qenv` class as input.
- Subsequently, we adopted a more general approach where the input is assumed to be character. This allows passing `teal_data` with code in `teal_data@code` slot, leading to the migration of this functionality to the `teal.data` ([PR #194](https://github.com/insightsengineering/teal.data/pull/194)).

# Implementation Plan

**Code Parser** consists of:

- **Code Graph** - The `code_dependency` / `code_graph` function constructs the structure of dependencies between objects and their occurrences in specific calls derived from the code.
- **Graph Parser** - The `get_code_dependency` / `get_object_code` function take a code graph as input and the names of objects existing in the code. It returns the code, including all necessary dependencies, to recreate the specified object and its influencers.

## Pseudo code / algorithm

<details><summary>Code Graph</summary>

- Take `code` as an input (a `character` or an `expression` with `srcref` attribute.
- Put the code in `utils::getParseData` to extract information about the parsed code with built-in functions.
- `utils::getParseData` creates a `data.frame` structure (`pd`) enumerating each call, and enumerating each object/symbol within calls. Each object/symbol has a `token` metadata specifying how it's treated by R (e.g., `SYMBOL`, `ASSIGNMENT OPERATOR`, `FUNCTION_CALL`, `SYMBOL_FORMALS` etc).
- Thanks to `pd`, we are able to bind all elements of all calls of the input code into a list. The list has a length equal to the number of calls (`calls_pd`).
- Then within calls, we would like to extract objects by their metadata (included in `token`) so that we seek for `"SYMBOL", "SYMBOL_FUNCTION_CALL"` and grep for `ASSIGN` operators to understand which object is influenced by other objects in this call.
- We also check for `COMMENT` tokens that contain `@linksto` tag to understand whether some calls should be assigned as influencers of other objects.
- With the above information, we need to build a structure that in some way presents:
    - which objects exist in which calls
    - which objects influence other objects and in which calls
    - which side-effects influence which objects in which calls
 
 :warning: the above structure and its creation is a part of this PR which simplifies the current implementation
</details>

<details><summary>Graph Parser</summary>

Having the **Code Graph** we take an object name (of multiple `names`) and we
- (1) seek for a call in which this object was created (let's call it call X)
- (2) we limit the input `calls_pd` of all calls until call X (let's call it `calls_pd_x`)
- (3) we identify all influencers and side effects of object in the `calls_pd_x`
- we repeat (1-3) for influencers from (3) and new `calls_pd_x` until all considered objects no longer have influencers

:warning: the above process simplification is also a part of this PR since side_effects are detected by Code Parser and could be detected by Code Graph
:warning: the above process will also be simplified in this PR as it is based on object names and calls indexes, but could be merged into a process that uses one of those two
</details>

# Notes 

- The relationship between objects is assumed to be conveyed through `<-`, `=`, or `->` assignment operators. No other object creation methods (such as `assign`, `<<-`, or any non-standard-evaluation method) are supported. This is addressed by using the `# @linktso` tag. 
- We do not assume any non-standard operations nor evaluations in data processing code; however, if someone needs to create objects the other way than with the assignment operators, the `# @linktso` tag is meant for it.
- Any specific side effects that should be returned with a specific object should be tagged with `# @linktso` tag at the end of the line where the side effect is created.


m7pr:
Pushed a small alternative https://github.com/insightsengineering/teal.data/pull/201/commits/3224a487397683a3bdb1543ffa5131325108bdc5 but this is not finished yet

R/utils-get_code_dependency.R
```diff
@@ -0,0 +1,498 @@
...
+  }
+}
+
+#' @keywords internal
+#' @noMd
+is_empty <- function(code){
+  identical(code, character(0)) || identical(trimws(code), "")
+}
+
+# # # examples and test -----------------------------------------------------------------------------------------------
```
gogonzo, line 299: please remove commented text

R/utils-get_code_dependency.R
```diff
@@ -0,0 +1,498 @@
...
+#' @param names A `character(n)` with object names.
+#'
+#' @return `character` vector of elements of `code` calls that were required to build the side-effects and
+#' influencing objects having and impact on the `object`
+#'
+#' @keywords internal
+get_code_dependency <- function(code, names) {
+  assert_classes(code, names)
+
+  if (is_empty(code)) return(code)
```
gogonzo, line 20: function used only once, don't need to create sugar in my opinion

```suggestion
  if (identical(code, character(0)) || identical(trimws(code), "")) return(code)
```

R/utils-get_code_dependency.R
```diff
@@ -0,0 +1,498 @@
...
+  assert_names(names, pd)
+
+  calls_pd <- extract_calls(pd)
+
+  graph <- code_graph(calls_pd)
+  indexes <- unlist(lapply(names, function(x) graph_parser(x, graph)))
+  as.character(code[unique(indexes)])
+}
+
+#' Group the result of `utils::getParseData()` into separate calls
```
gogonzo, line 33:
```suggestion
#' Split the result of `utils::getParseData()` into separate calls
```

R/utils-get_code_dependency.R
```diff
@@ -0,0 +1,498 @@
...
+#' @param pd (`data.frame`) A result of `utils::getParseData()`.
```
gogonzo, line 34:
```suggestion
#' @param pd (`data.frame`) A result of [utils::getParseData()] limited to three columns (`"token"`, `"text"`, `"id"`)
```

R/utils-get_code_dependency.R
```diff
@@ -0,0 +1,498 @@
...
+      } else {
+        x[sym_cond, "text"]
+      }
+
+      ### NOTE(3): What if there are 2+ assignments, e.g. a <- b -> c or e.g. a <- b <- c.
+      ### NOTE(2): For cases like 'eval(expression(b <- b + 2))' removes 'eval(expression('.
+      ### NOTE(1): Cases like 'data(iris)' that do not have an assignment operator.
+    }
+  )
+}
```
gogonzo, line 167: is it needed 
```suggestion
```

R/utils-get_code_dependency.R
```diff
@@ -0,0 +1,498 @@
...
+
+  graph <- mapply(c, side_effects, cooccurence, SIMPLIFY = FALSE)
+
+  remove_graph_duplicates(graph)
+
+}
+
+#' @keywords internal
+#' @noRd
+extract_occurence <- function(calls_pd) {
```
gogonzo, line 96: Please write docs of this function - what technically occurrence here means
```suggestion
#' @keywords internal
#' @noRd
extract_occurrence <- function(calls_pd) {
```

R/utils-get_code_dependency.R
```diff
@@ -0,0 +1,498 @@
...
+      ### NOTE(3): What if there are 2+ assignments, e.g. a <- b -> c or e.g. a <- b <- c.
+      ### NOTE(2): For cases like 'eval(expression(b <- b + 2))' removes 'eval(expression('.
+      ### NOTE(1): Cases like 'data(iris)' that do not have an assignment operator.
+    }
+  )
+}
+
+#' @keywords internal
+#' @noRd
+extract_side_effects <- function(calls_pd) {
```
gogonzo, line 171: please expand docs of this function - please explain side effects term

gogonzo (commented):
First quick review - Please remove commented code and elaborate more about functions I mentioned in a comment. I also get R CMD check test failure. Please confirm if you run into the same ~ `get_code(datanames) no such argument "datanames"`

R/utils-get_code_dependency.R
m7pr: Yes, this will be removed before the merge : ) It's mainly here for easier review

m7pr:
> I also get R CMD check test failure. Please confirm if you run into the same ~ get_code(datanames) no such argument "datanames"

I think the branch name should be changed, so it's visible it's branched out of `@refactor` branch?

R/utils-get_code_dependency.R
m7pr: oh yes, this is the body of `extract_occurrence` : P highly needed and requires simplification, but not deletion haha!

R/utils-get_code_dependency.R
m7pr: I think the result of this function is limited, but not the input parameter. Maybe we can move this to `@return`/`Value` section : )

R/utils-get_code_dependency.R
gogonzo: I don't know whether this is easier. I needed to scroll down and figure out what `is_empty` means. And it means something else than `rlang` equivalent.

```r
rlang::is_empty("")
# [1] FALSE
```

R/utils-get_code_dependency.R
gogonzo: sorry my bad, i though it was outside of the function

R/utils-get_code_dependency.R
gogonzo: true, my mistake

R/utils-get_code_dependency.R
m7pr: I do select the name and press ctrl/fn+F2 in RStudio and it redirects you to the function definition. IMHO there is less clutter reviewing and you can focus on major parts of the implementation during the review and do not necessarily need to look at assertions that are just long. For sure I will brought this back to `get_code_dependencies`

R/utils-get_code_dependency.R
m7pr: no worries, c05f27f382473b3abb89a11b61b5ac581ed5e843

R/utils-get_code_dependency.R
m7pr: 03710f3e41b27f47e0d2fcbf0b5650df5297ef70

m7pr:
The R CMD Check was fixed with https://github.com/insightsengineering/teal.data/pull/201/commits/6a4b5f2c3569f8e47263a99a390a0723c1fe6893 for which I also created a separate issue in `teal.code` https://github.com/insightsengineering/teal.code/issues/178

R/utils-get_code_dependency.R
gogonzo: > IMHO there is less clutter reviewing and you can focus on major parts

Yes, but you have also extra code to maintain the package. I understand the reason but I disagree. I think you made a great job by splitting former functions into meaningful parts but these utilities is little too much in my opinion. When analysing the code you need to know what each line does. Here, instead of knowing "what is the assertion on the code" one needs to go to the private util to read two lines of code. So not a great simplification of the code - one vs two.

`assert_classes` is wrong for many reasons. Name is wrong, as it looks more like `assert_get_code_dependency_formals`. `assert_classes(code, names)` sounds like `checkmate::assert_class` but it is something completely different. It doesn't fit to any `assert_` function.

I've already highlighted that `is_empty` is exactly the same name as `rlang::is_empty` function but return different results. `rlang` is by the way used in the package so I had reasonable suspicious that this is `rlang` function.

R/utils-get_code_dependency.R
```diff
@@ -0,0 +1,530 @@
...
+  }
+
+}
+
+# utils -----------------------------------------------------------------------------------------------------------
+
+#' @keywords internal
+#' @noMd
+assert_classes <- function(code, names) {
+  checkmate::assert_multi_class(code, classes = c("character", "expression"))
```
gogonzo, line 292: `get_code_dependency` is used once on `@code` which is always a character
```suggestion
  checkmate::assert_class(code, classes = "character")
```

R/utils-get_code_dependency.R
```diff
@@ -0,0 +1,530 @@
...
+#' @return `character` vector of elements of `code` calls that were required to build the side-effects and
+#' influencing objects having and impact on the `object`
+#'
+#' @keywords internal
+get_code_dependency <- function(code, names) {
+  assert_classes(code, names)
+
+  if (is_empty(code)) return(code)
+
+  code <- assert_code(code) # Turns code into expression with srcref attribute.
```
gogonzo, line 22: `assert_code` is an imprecise name for this function. Besides, code can't be an expression, we know it is always a character, so you can easily replace it with `parse`  

```suggestion
  code <- parse(code, keep.source = TRUE) 
```

R/utils-get_code_dependency.R
```diff
@@ -0,0 +1,530 @@
...
+
+          ans <- if (text[1] == "->") { # NOTE(3)
+            rev(text[-1])
+          } else {
+            text[-1]
+          }
+          c(ans[1], ":", unique(ans[-1]))
+        } else {
+          x[sym_cond, "text"]
+        }
```
gogonzo, line 174: `text` and `ans` could be omitted and code could be simpler.

```suggestion
        if (length(ass_cond)) { # NOTE(1)
          sym_cond <- sym_cond[sym_cond > ass_cond] # NOTE(2)
          if (x$text[ass_cond] == "->") { # NOTE(3)
            sym_cond <- rev(sym_cond)
          } 
          c(x[sym_cond[1], "text"], ":", unique(x[-sym_cond[1], "text"]))
        } else {
          x[sym_cond, "text"]
        }
```

R/utils-get_code_dependency.R
m7pr: Yes, I see it didnt bring any help. I did mention in my first comment, when asking for a review, that I am not planning to keep those ever https://github.com/insightsengineering/teal.data/pull/201#issuecomment-1827990169 Will remove those utils as it only causes unnecessary confusion!

R/utils-get_code_dependency.R
m7pr: oh yeah, this a remaining from the past where this could be used by the end user but not the case anymore. trimming to character only

R/utils-get_code_dependency.R
m7pr: Sorry for the confusion - just removed 7402a4dc752cda2e6e7c79f0f8f32f8c78d2cf02 !

R/utils-get_code_dependency.R
m7pr: needed to read it a couple of times! thanks, this is nice, will verify in a few minutes

m7pr: Cool, just reviewed against tests. This is great, thanks. Merging

R/utils-get_code_dependency.R
```diff
@@ -0,0 +1,491 @@
...
+          if (x$text[ass_cond] == "->") { # NOTE(3)
+            sym_cond <- rev(sym_cond)
+          }
+          append(x[sym_cond, "text"], ":", after = 1)
+        } else {
+          x[sym_cond, "text"]
+        }
+      } else {
+        x[sym_cond, "text"]
+      }
```
gogonzo, line 182: Look at the suggestion below. Imagine `ass_cond` could be an index of the `assing()` function token and we could handle `assing()` ;]

```suggestion
      ass_cond <- grep("ASSIGN", x$token)

      # If there was an assignment operation detect direction of it.
      if (length(ass_cond) > 0) { # NOTE(1)
        sym_cond <- sym_cond[sym_cond > ass_cond] # NOTE(2)
        if (any(x$text[ass_cond] == "->")) { # NOTE(3)
          sym_cond <- rev(sym_cond)
        }
      }

      if (length(sym_cond) > 1) {
        append(x[sym_cond, "text"], ":", after = 1)
      } else {
        x[sym_cond, "text"]
      }
```

m7pr: This preserves the functionality and all tests pass. I would merge as this is a proper simplification.
I'm just not sure about `any(x$text[ass_cond] == "->")` as I don't think this (and mine) implementation handles calls with multiple assignments like `a <- 5 -> b `.

What do you mean about handling `assign()` function? How would that work?

R/utils-get_code_dependency.R
```diff
@@ -0,0 +1,491 @@
...
+    names(symbols_table[symbols_table == 2])
+  }
+
+  functions <- extract_function_names(calls_pd)
+
+  lapply(
+    calls_pd,
+    function(x) {
+      sym_cond <- which(x$token == "SYMBOL" | (x$token == "SYMBOL_FUNCTION_CALL" & x$text %in% functions))
+      sym_cond <- sym_cond[!used_in_function(x, sym_cond)]
```
gogonzo, line 157: Please take `get_children` from `extract_calls` and you'll have a nice simplification

```suggestion
  is_in_function <- function(x) {
    # If an object is a function parameter,
    # then in calls_pd there is a `SYMBOL_FORMALS` entry for that object.
    function_id <- x[x$token == "FUNCTION", "parent"]
    if (length(function_id)) {
      x$id %in% get_children(x, function_id)$id
    } else {
      rep(FALSE, nrow(x))
    }
  }

  lapply(
    calls_pd,
    function(call_pd) {
      # what occurs in a function body is not tracked
      x <- call_pd[!is_in_function(call_pd), ]
      sym_cond <- which(x$token %in% c("SYMBOL", "SYMBOL_FUNCTION_CALL"))
```

gogonzo (commented):
Please run `styler::style_pkg("your/local/teal.data ")`.

chlebowa:
This is a little off-topic but `get_code_dependency` can raise a warning that reads `Objects not found in 'qenv' environment: (...)`. Could you change this to read `teal_data` instead of `qenv`?

R/utils-get_code_dependency.R
m7pr: https://github.com/insightsengineering/teal.data/pull/201/commits/3480f9847b487553402a576fc18e62994d028514

m7pr:
@chlebowa I think you are looking at `main` branch. Such warning does not exist in this feature branch

R/utils-get_code_dependency.R
m7pr: https://github.com/insightsengineering/teal.data/pull/201/commits/382da4555306cf320539f67c4317491798d09726

m7pr:
Hey, here is the code you can use for testing. @gogonzo it's ready for one more round of reviews.

```r
# # examples and test -----------------------------------------------------------------------------------------------
# TOBE REMOVED BEFORE THE MERGE
code <- "

  a <- 5
  b <- a + 3
  a <- a + 6
  5 -> c # @linksto a

"

graph_expected <- list(
  c("a"),
  c("b", ":", "a"),
  c("a", ":", "a"),
  c("a", "c")
)

code_a_expected <- c("a <- 5", "a <- a + 6", "c <- 5")
code_b_expected <- c("a <- 5", "b <- a + 3")
code_c_expected <- c("c <- 5")


code_2 <- "

  a <- 1
  b <- identity(x = a)
  a <- 2

"

graph_expected_2 <- list(
  c("a"),
  c("b", ":", "identity", "a"),
  c("a")
)

code_b_expected_2 <- c("a <- 1", "b <- identity(x = a)")

code_3 <- "

  a <- 1
  assign('b', 5) # @linksto b
  b <- b + 2

"

graph_expected_3 <- list(
  c("a"),
  c("b", "assign"),
  c("b", ":", "b")
)

code_b_expected_3 <- c("assign(\"b\", 5)", "b <- b + 2")

code_4 <- "
  iris[1:5, ] -> iris2
  iris_head <- head(iris) # @linksto iris3
  iris3 <- iris_head[1, ] # @linksto iris2
  classes <- lapply(iris2, class)
"

graph_expected_4 <- list(
  c("iris2", ":", "iris"),
  c("iris3", "iris_head", ":", "head", "iris"),
  c("iris2", "iris3", ":", "iris_head"),
  c("classes", ":", "lapply", "iris2", "class")
)

code_c_expected_4 <- c(
  "iris2 <- iris[1:5, ]",
  "iris_head <- head(iris)",
  "iris3 <- iris_head[1, ]",
  "classes <- lapply(iris2, class)"
)

make_graph <- function(code) {
  pd <- utils::getParseData(code)
  calls_pd <- extract_calls(pd)
  code_graph(calls_pd)
}

parsed_code <- parse(text = code, keep.source = TRUE)
graph <- make_graph(parsed_code)

parsed_code_2 <- parse(text = code_2, keep.source = TRUE)
graph_2 <- make_graph(parsed_code_2)

parsed_code_3 <- parse(text = code_3, keep.source = TRUE)
graph_3 <- make_graph(parsed_code_3)

parsed_code_4 <- parse(text = code_4, keep.source = TRUE)
graph_4 <- make_graph(parsed_code_4)

testthat::test_that("code_graph returns proper structure of the dependency graph", {

  testthat::expect_identical(
    graph,
    graph_expected
  )

  testthat::expect_identical(
    graph_2,
    graph_expected_2
  )

  testthat::expect_identical(
    graph_3,
    graph_expected_3
  )

  testthat::expect_identical(
    graph_4,
    graph_expected_4
  )

})

testthat::test_that("graph_parser returns proper code based on code_graph", {

  names <- 'a'
  indexes <- unlist(lapply(names, function(x) graph_parser(x, graph)))
  testthat::expect_identical(
    as.character(parsed_code[indexes]),
    code_a_expected
  )

  names <- 'b'
  indexes <- unlist(lapply(names, function(x) graph_parser(x, graph_2)))
  testthat::expect_identical(
    as.character(parsed_code_2[indexes]),
    code_b_expected_2
  )

  names <- 'b'
  indexes <- unlist(lapply(names, function(x) graph_parser(x, graph_3)))
  testthat::expect_identical(
    as.character(parsed_code_3[indexes]),
    code_b_expected_3
  )

  names <- 'classes'
  indexes <- unlist(lapply(names, function(x) graph_parser(x, graph_4)))
  testthat::expect_identical(
    as.character(parsed_code_4[indexes]),
    code_c_expected_4
  )

})

testthat::test_that("get_code_dependency returns proper code", {

  testthat::expect_identical(
    get_code_dependency(code, names = 'a'),
    code_a_expected
  )

  testthat::expect_identical(
    get_code_dependency(code, names = 'b'),
    code_b_expected
  )

  testthat::expect_identical(
    get_code_dependency(code, names = 'c'),
    code_c_expected
  )

  testthat::expect_identical(
    get_code_dependency(code_2, names = 'b'),
    code_b_expected_2
  )

  testthat::expect_identical(
    get_code_dependency(code_3, names = 'b'),
    code_b_expected_3
  )

  testthat::expect_identical(
    get_code_dependency(code_4, names = 'classes'),
    code_c_expected_4
  )

})

testthat::test_that("remove_graph_duplicates cleans graph out ouf duplicates", {
  graph <- list(
    c("a", "a", ":", "b"),
    c("a", ":", "b", "b"),
    c("a", ":", "a")
  )
  testthat::expect_equal(
    remove_graph_duplicates(graph),
    list(
      c("a", ":", "b"),
      c("a", ":", "b"),
      c("a", ":", "a")
    )
  )

})

```

tests/testthat/test-get_code.R
```diff
@@ -0,0 +1,476 @@
...
+    "a <- 1",
+    "assign('b', 5)",
+    "b <- b + 2"
+  )
+  tdata <- eval_code(teal_data(), code)
+  datanames(tdata) <- c("a", "b")
+  testthat::expect_error(
+    get_code(tdata, datanames = "b"),
+    "b <- b + 2"
+  )
```
gogonzo, line 103: It won't throw error if we support this. Suggested change is an expectation
```suggestion
  testthat::expect_identical(
    get_code(tdata, datanames = "b"),
    c("assign('b', 5)", "b <- b + 2")
  )
```

gogonzo (approved):
Great! 💪

R/utils-get_code_dependency.R
```diff
@@ -0,0 +1,267 @@
...
+      rep(FALSE, nrow(x))
+    }
+  }
+  lapply(
+    calls_pd,
+    function(call_pd) {
+
+      # What occurs in a function body is not tracked.
+      x <- call_pd[!is_in_function(call_pd), ]
+      sym_cond <- which(x$token %in% c("SYMBOL", "SYMBOL_FUNCTION_CALL"))
```
chlebowa, line 141: `which` returns `integer(0)` if `x` is all `FALSE`. Please make sure this will not kick us in the backside somewhere.

R/utils-get_code_dependency.R
```diff
@@ -0,0 +1,267 @@
...
+        }
+      })
+    )
+  influencers <- setdiff(influencers, skip)
+
+  if (length(influencers)) {
+    influencers_ids <-
+      lapply(influencers, function(influencer) {
+        graph_parser(influencer, graph[1:max(which(occurence))], skip)
+      })
```
chlebowa, line 261: Is `skip` indispensible?
```suggestion
      lapply(setdiff(influencers, x), function(influencer) {
        graph_parser(influencer, graph[1:max(which(occurence))])
      })
```

R/utils-get_code_dependency.R
```diff
@@ -0,0 +1,267 @@
...
+graph_parser <- function(x, graph, skip = NULL) {
+  skip <- c(x, skip)
+
+  occurence <-
+    vapply(
+      graph, function(call) {
+        if (":" %in% call) {
+          call <- call[1:(which(":" == call) - 1)]
+        }
+        x %in% call
```
chlebowa, line 242:
```suggestion
        ind <- match(":", call, nomatch = length(call) +1L)
        x %in% call[seq_len(ind - 1L)]
```

R/utils-get_code_dependency.R
```diff
@@ -0,0 +1,267 @@
...
+      },
+      logical(1)
+    )
+
+  influencers <-
+    unlist(
+      lapply(graph[occurence], function(call) {
+        if (":" %in% call) {
+          call[(which(":" == call) + 1):length(call)]
+        }
```
chlebowa, line 252:
```suggestion
        ind <- match(":", call, nomatch = 0L)
        x %in% call[seq_len(ind + 1L)]
```

R/utils-get_code_dependency.R
```diff
@@ -0,0 +1,267 @@
...
+      })
+    )
+  influencers <- setdiff(influencers, skip)
+
+  if (length(influencers)) {
+    influencers_ids <-
+      lapply(influencers, function(influencer) {
+        graph_parser(influencer, graph[1:max(which(occurence))], skip)
```
chlebowa, line 260: `any(occurence)` better be `TRUE`

chlebowa (dismissed):
Documentation needs work.
Some suggestions for code but nothing crucial.

R/teal_data-get_code.R
```diff
@@ -6,6 +6,8 @@
 #' Use `datanames` to limit the code to one or more of the data sets enumerated in `@datanames`.
 #' If the code has not passed verification, a warning will be prepended.
 #'
+#' @inheritSection get_code_dependency Notes
```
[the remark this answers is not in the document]
m7pr: Cool, will rethink

R/utils-get_code_dependency.R
```diff
@@ -0,0 +1,267 @@
...
+
+#' Return the lines of code (with side-effects) needed to reproduce the object
+#'
+#' @section Notes:
+#' This function assumes that object relationships are established using the `<-`, `=`, or `->` assignment
+#' operators. It does not support other object creation methods like `assign` or `<<-`, nor non-standard-evaluation
+#' methods. To specify relationships between side-effects and objects, you can use the comment tag
+#' `# @linksto object_name` at the end of a line where the side-effect occurs.
+#'
+#' @param code An `expression` with `srcref` attribute or a `character` with the code.
```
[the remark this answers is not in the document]
m7pr: oh yeah

R/utils-get_code_dependency.R
```diff
@@ -0,0 +1,267 @@
...
+  checkmate::assert_character(names)
+
+  if (identical(code, character(0)) || identical(trimws(code), "")) {
+    return(code)
+  }
+
+  code <- parse(text = code, keep.source = TRUE)
+  pd <- utils::getParseData(code)
+
+  # Assert names are actually in code.
```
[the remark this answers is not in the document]
m7pr: a1ace5504d21e70a582d21c3deae50ceb693efd1

R/utils-get_code_dependency.R
```diff
@@ -0,0 +1,267 @@
...
+    warning(
+      "Object(s) not found in code: ",
+      toString(setdiff(names, symbols))
+    )
+  }
+
+  calls_pd <- extract_calls(pd)
+
+  graph <- code_graph(calls_pd)
+  indexes <- unlist(lapply(names, function(x) graph_parser(x, graph)))
```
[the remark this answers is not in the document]
m7pr: d2872401b16bf0e51d233572720bea2866424e4f

R/utils-get_code_dependency.R
```diff
@@ -0,0 +1,267 @@
...
+  as.character(code[unique(indexes)])
+}
+
+#' Split the result of `utils::getParseData()` into separate calls
+#' @param pd (`data.frame`) A result of `utils::getParseData()`.
+#'
+#' @return A `list` of length equal to the number of calls in code used to create `pd`. Each element of the returned
+#' `list` is one call of a structure similar to `pd` (`utils::getParseData()`) but limited to three columns
+#' (`"token"`, `"text"`, `"id"`).
```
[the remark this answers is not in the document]
m7pr: yes, but I think we also return the 4th column `parent`. will include and fix after

R/utils-get_code_dependency.R
```diff
@@ -0,0 +1,267 @@
...
+  }
+}
+
+fix_comments <- function(calls) {
+  # If the first token is a COMMENT, then it belongs to the previous call.
+  if (length(calls) >= 2) {
+    for (i in 2:length(calls)) {
+      if (grepl("@linksto", calls[[i]][1, "text"])) {
+        calls[[i - 1]] <- rbind(calls[[i - 1]], calls[[i]][1, ])
+        calls[[i]] <- calls[[i]][-1, ]
```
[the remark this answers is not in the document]
m7pr: Not closing this conversation yet - check this out 507ff9d17698dd94180b1eeb95a4a8c85379e2c5
There is an improvement as merging lists happens on a new list, that gets extended during each iteration anyway. However the reduction of the old list (by the first row) remains.

R/utils-get_code_dependency.R
```diff
@@ -0,0 +1,267 @@
...
+
+      # If there was an assignment operation detect direction of it.
+      ass_cond <- grep("ASSIGN", x$token)
+      if (length(ass_cond)) { # NOTE(1)
+        sym_cond <- sym_cond[sym_cond > ass_cond] # NOTE(2)
+      }
+      if ((length(ass_cond) && x$text[ass_cond] == "->") || !length(ass_cond)) { # NOTE(3)
+          sym_cond <- rev(sym_cond)
+      }
+      append(x[sym_cond, "text"], ":", after = 1)
```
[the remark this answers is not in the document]
m7pr: Just because it's short and contains a symbol not allowed in object names, I personally think that `:` is easier to be found in a vector of actual names, like the one below:

```r
graph_expected_4 <- list(
  c("iris2", ":", "iris"),
  c("iris3", "iris_head", ":", "head", "iris"),
  c("iris2", "iris3", ":", "iris_head"),
  c("classes", ":", "lapply", "iris2", "class")
)
```

when compares to

```r
graph_expected_4 <- list(
  c("iris2", "affected by", "iris"),
  c("iris3", "iris_head", "affected by", "head", "iris"),
  c("iris2", "iris3", "affected by", "iris_head"),
  c("classes", "affected by", "lapply", "iris2", "class")
)
```

R/utils-get_code_dependency.R
```diff
@@ -0,0 +1,267 @@
...
+#' parsed code. It helps you understand which objects are needed to recreate a specific object.
+#'
+#' @param calls_pd A `list` of `data.frame`s, which is a result of `utils::getParseData()` grouped into separate calls.
+#' A result of `extract_calls()` function.
+#'
+#' @return A `list` (of length of input `calls_pd`) where each element represents one call. Each element consists of a
+#' character vector containing names of objects that were influenced by this call, and names of objects influencing this
+#' call. Influencers appear after `":"` string, e.g. `c("a", ":", "b", "c")` means a call influenced object named `a`,
+#' and the object was influenced by objects named `b` and `c`. If an object was marked with `@linksto` side effects tag
+#' then it appears as an influenced object at the beginning of a vector.
```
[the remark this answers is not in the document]
m7pr: yeah, I didn't like `influenced` either and was waiting to get something better. however `effector` sounds like it's something that got stimulated, which means it was already `influenced/triggered`.

I will substitute `influenced` with `affected`, but will think a bit about the `effector`.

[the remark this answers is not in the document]
m7pr: 11f84555174bf04fc17ac40886d179619f1097e3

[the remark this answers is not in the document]
chlebowa: An effector is something that exerts an effect. I understand it may not be intuitive.

R/utils-get_code_dependency.R
```diff
@@ -0,0 +1,267 @@
...
+    }
+  )
+}
+
+#' @keywords internal
+#' @noRd
+remove_graph_duplicates <- function(graph) {
+  # This function is just a nice to have.
+  # This is not crucial for the implementation.
+  # Purpose: remove duplicates before ":" and after, in each element of list.
```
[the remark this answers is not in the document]
m7pr: This happens on concatenation of `side_effects` and `cooccurrence` in 

`graph <- mapply(c, side_effects, cooccurence, SIMPLIFY = FALSE)`

and duplicates can appear because there could be a line like `a <- fun1(b) # @linksto a` , which means `a` was in `side_effects` and `cooccurrence`

I think we could reduce duplicates in the `cooccurence` part for cases like `a <- fun1(b, b, b)` with this 6347792e7394e4bf8dce5d1b9a58def97e072e07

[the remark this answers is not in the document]
chlebowa: OR play with the `FUN` in `mapply` a little?

R/utils-get_code_dependency.R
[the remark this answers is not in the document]
m7pr: sorry forgot about b8e272060575910e07151d184acdd13094cf59cf

I think currently it is

```r
fix_comments <- function(calls) {
  # If the first token is a COMMENT, then it belongs to the previous call.

  if (length(calls) >= 2) {
    calls_new <- list()
    calls_new[[1]] <- calls[[1]]
    for (i in 2:length(calls)) {
      if (grepl("@linksto", calls[[i]][1, "text"])) {
        calls_new[[i - 1]] <- rbind(calls[[i - 1]], calls[[i]][1, ])
        calls[[i]] <- calls[[i]][-1, ]
      } else {
        calls_new[[i]] <- calls[[i]]
      }
    }
  }
  calls_new
}
```

I not convinced it's worth it

R/utils-get_code_dependency.R
m7pr: yeah, if `sym_cond` is integer(0)  which means there were no symbols in the call then it's ok. Check this test f472e68e798008aac4a8f62a137f4fc41221230d

m7pr: also added a return statement if sym_cond is empty to fail earlier 3b53e80f044c83c0f9590c5ac6abe6b20e26ba81

R/utils-get_code_dependency.R
m7pr: yes, because it keeps getting extended in every loop.

```r
a <- 1
b <- a
c <- b
a <- c
```

and if you don't have skip, then you would end up in the infinite loop

R/utils-get_code_dependency.R
m7pr: nice, thanks

R/utils-get_code_dependency.R
chlebowa: And the `setdiff` is not enough?

R/utils-get_code_dependency.R
```diff
@@ -0,0 +1,267 @@
+# get_code_dependency ----
+
+#' Return the lines of code (with side-effects) needed to reproduce the object
+#'
+#' @section Notes:
+#' This function assumes that object relationships are established using the `<-`, `=`, or `->` assignment
+#' operators. It does not support other object creation methods like `assign` or `<<-`, nor non-standard-evaluation
+#' methods. To specify relationships between side-effects and objects, you can use the comment tag
+#' `# @linksto object_name` at the end of a line where the side-effect occurs.
```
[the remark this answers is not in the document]
m7pr: changed the section name from notes to  details in here 543e302b02b1fb22574671fb1cb82c768e8975d5

R/teal_data-get_code.R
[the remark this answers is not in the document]
m7pr: Alrighty, I added a simple note about possibility to read about `get_code_dependency` d8a1eaae54bb9b82746eec10d695509c5c29681f

R/utils-get_code_dependency.R
```diff
@@ -0,0 +1,267 @@
...
+#'
+#' @param code An `expression` with `srcref` attribute or a `character` with the code.
+#' @param names A `character(n)` with object names.
+#'
+#' @return `character` vector of elements of `code` calls that were required to build the side-effects and
+#' influencing objects having and impact on the `object`
```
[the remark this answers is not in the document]
m7pr: TBD

R/utils-get_code_dependency.R
m7pr: I actually think this needs to be 

```r
ind <- match("<-", call, nomatch = 0L)
call[(ind + 1L):length(call)]
```

m7pr: I will leave it as 

```r
       if ("<-" %in% call) {
          ind <- match("<-", call, nomatch = 0L)
          call[(ind + 1L):length(call)]
        }
```

chlebowa: That's what the `nomatch` is for.

I don't know what names you mean.

m7pr: `call` is a vector of names. if there is no `:` nor `<-` and call looks like this `c("a")` then there are no influencers and I rather return `character(0)` and not TRUE/FALSE

m7pr: the current statement that I think works is here 91327a9bc1592a4d58b104d5f71d4d9b15909121

R/utils-get_code_dependency.R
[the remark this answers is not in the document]
m7pr: 91327a9bc1592a4d58b104d5f71d4d9b15909121

R/utils-get_code_dependency.R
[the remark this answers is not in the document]
m7pr: yeah, could be way simpler than dealing with duplicates!

R/utils-get_code_dependency.R
chlebowa: Ooooh, I see the problem.
```
        ind <- match(":", call, nomatch = 0L)
        x %in% call[seq_len(ind + 1L)]
```

I thought you were talking about the "<-" and not the `seq_len`.

You're right, it should be `call[(ind + 1L):length(call)]`.

R/utils-get_code_dependency.R
m7pr: Hmm, let's consider below flow

We seek for code of `a`. 
1st iteration: 'x <- "a"` and `influencers <- "c"` and then we run `graph_parser` recursively for `x <- "c"` and `graph` is reduced to `1:4` (as `a` exists in line 4)
2nd iteration: `x <- "c"` and `influencers <- "b"` and then we run `graph_parser` recursively for `x <- "b"` and `graph` is 
 reduced to `1:4` (as `c` exists in line 4)
3rd iteration: `x <- "b"` and `influencers <- "a"` and again you run `graph_parser` recursively for `x <- "a"` and `graph` is reduced to `1:3` (as `b` exists in line 3)

Ok, so I think we no longer arrive at `a` being influenced by `c` again since the graph got reduced. Will try with `setdiff`

m7pr: Thanks for the valuable feedback - just removed `skip` parameter 0891a1f8c52ced3a81984add5a95a849d8876ba5 and added a test that detects if we fall in a loop

R/utils-get_code_dependency.R
m7pr: bb5bb39c0c77e365ea8d1ceeb62feb3a4cd87db4

R/utils-get_code_dependency.R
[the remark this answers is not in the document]
m7pr: bringing back the simpler code as it's very rare and data.frames are small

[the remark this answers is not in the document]
m7pr: https://github.com/insightsengineering/teal.data/commit/07a000b4375d746c16ff9f98640f12009d65c882

R/utils-get_code_dependency.R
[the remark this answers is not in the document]
m7pr: Hey @chlebowa the length of `names` does not have an effect of length of the restulting `character` vector. 548b3b0f39cc8ba1df7a3086a8bf1fc41670c6be

It's gonna be always lenght 1, but it will have more calls collapsed via '\n'

gogonzo:
Good job @m7pr @chlebowa and Me :D

m7pr:
Hey @chlebowa thanks for the final review and a huge documentation cleanup! You are da man! Thanks @gogonzo for all the feedback related to implementation. It looks like code parser is way simpler than what we had in the first attempt.

PR #214, opened 2023-12-11
213 upgrade `teal.code` to `Depends`

### LINKED ISSUE #213: upgrade `teal.code` to `Depends`
Since some methods from `teal.code` are used in `teal.data`, the entire `teal.code` namespace should be re-exported or attached in `teal.data`.

After some discussion we decided to attach, so `teal.code` will be moved from `Imports` to `Depends`.

chlebowa:
Fixes #213


gogonzo (commented):
👍

gogonzo (approved):

PR #215, opened 2023-12-11
`join_key` can set default parent

### LINKED ISSUE #203: [Feature Request]: Set default parent
### Problem:

When creating a new `join_keys` object with foreign keys it requires an external call to define parents

``` r
library(teal.data)
join_keys(
  join_key("A", keys = "id_a"),
  join_key("B", keys = "id_b"),
  join_key("A", "B", keys = c(id_b = "id_a"))
) |> parents()
#> list()
```

### Benefit

Single calls create valid/complete `join_keys` without any extra manipulation.

### Solution A: assume directonality on `join_key()` call 

- `join_key("A", "B", keys = c(id_b = "id_a"))` assumes first/second dataset to be parent
  - add parameter paremeter to disable this assumption
  - alternative, don't assume and add parameter to define parent
    - `join_key("A", "B", keys = c(id_b = "id_a"), parent = "A")`

### Solution B: add parent argument to constructor

Only valid for constructor

```r
data_1 <- data.frame(X = factor(1:10), Y = 21:30, Z = 1:10)
data_2 <- data.frame(W = factor(10:1), V = factor(5:14), M = rep(1:5, 2))

data <- teal_data(
  D1 = data_1,
  D2 = data_2,
  join_keys = join_keys(
    join_key("D1", "D2", c("X" = "W")),
    parents = list(D2 = "D1")
  )
)
```

<details>

<summary>Draft implementation of solution 2</summary>

```diff
diff --git a/R/join_keys.R b/R/join_keys.R
index 8b962c07..7cf36105 100644
--- a/R/join_keys.R
+++ b/R/join_keys.R
@@ -62,7 +62,16 @@ join_keys <- function(...) {
 #' @order 1
 #' @export
 join_keys.default <- function(...) {
-  c(new_join_keys(), ...)
+  x <- rlang::list2(...)
+  new_parents <- list()
+  if (!is.null(names(x)) && "parents" %in% names(x)) {
+    new_parents <- x[["parents"]]
+    checkmate::assert_list(new_parents, names = "named", types = "character", .var.name = "parents")
+    x <- x[names(x) != "parents"]
+  }
+  new_jk <- do.call(c, append(list(new_join_keys()), x))
+  parents(new_jk) <- new_parents
+  new_jk
 }
```

</details>

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [X] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [X] I agree to follow this project's Security Policy.

averissimo:
# Pull Request

Fixes #203

### Changes description / TODO while wip

- [x] Add parameter on `join_key()` and `[<-.join_keys` that determines which dataset is parent
- [x] Protects against cyclical parents when creating or merging `join_keys`
- Tests
  - [x] Corrects existing tests to include new logic
  - [x] Re-write / add tests to account for usage
- Update documentation
  - [x] Man pages
  - [x] Update `join_key` vignette

### 🗨️ Discussion

- Q: should `jk[[i]][[j]] <- key` define a parent relationship?
  - consider that `jk[[i]] <- list(j = key, m = key2)` would have to set parents for `j` and `m`
    - might also have to add `parent` parameter.
  - My intuition is that `[[` should be left undirected


R/join_keys-c.R
```diff
@@ -28,13 +28,13 @@ c.join_keys <- function(...) {
     f = function(.x, .y) {
       assert_compatible_keys2(.x, .y)
       out <- utils::modifyList(.x, .y, keep.null = FALSE)
-      attr(out, "parents") <- .merge_parents(.x, .y)
+      parents(out) <- .merge_parents(.x, .y)
```
averissimo, line 31: Validation for acyciclic graphs

R/join_key.R
```diff
@@ -2,30 +2,35 @@
...
 #'
 #' @export
 #'
 #' @examples
 #' join_key("d1", "d2", c("A"))
 #' join_key("d1", "d2", c("A" = "B"))
 #' join_key("d1", "d2", c("A" = "B", "C"))
-join_key <- function(dataset_1, dataset_2 = dataset_1, keys) {
+join_key <- function(dataset_1, dataset_2 = dataset_1, keys, parent = c("dataset_1", "dataset_2", "none")) {
+  parent <- checkmate::matchArg(parent, choices = c("dataset_1", "dataset_2", "none"), .var.name = "parent")
```
averissimo, line 33: Last argument is needed. Upstream issue https://github.com/mllg/checkmate/issues/251

R/join_key.R
```diff
@@ -25,7 +28,8 @@
 #' join_key("d1", "d2", c("A"))
 #' join_key("d1", "d2", c("A" = "B"))
 #' join_key("d1", "d2", c("A" = "B", "C"))
-join_key <- function(dataset_1, dataset_2 = dataset_1, keys, parent = "dataset_1") {
+join_key <- function(dataset_1, dataset_2 = dataset_1, keys, parent = c("dataset_1", "dataset_2", "none")) {
+  parent <- tryCatch(match.arg(parent), error = function(err) parent)
   checkmate::assert_choice(parent, choices = c("dataset_1", "dataset_2", "none"))
```
chlebowa: Why not just `  parent <-match.arg(parent)`?

R/join_key.R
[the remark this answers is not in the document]
averissimo: The only benefit is the consistent message, but the function is buggy and incomplete :-\

I could live with repeated character verctor or using `eval(formals(join_key)[["parent"]])`, but when submitting the PR upstream I found yet another problem.

Maybe we can rethink when https://github.com/mllg/checkmate/pull/252 (or equivalent) is merged.

I'm reverting back to `match.arg` :+1:

R/join_key.R
```diff
@@ -2,30 +2,35 @@
...
+#' where `names(keys)` maps columns in `dataset_1` corresponding to columns of
+#' `dataset_2` given by the elements of `keys`.
 #'
 #' If unnamed, the same column names are used for both datasets.
 #'
 #' If any element of the `keys` vector is empty with a non-empty name, then the name is
 #' used for both datasets.
+#' @param parent (`character(1)`) Indicates the parent dataset in a parent-child
+#' relationship or `none` if it is an undirected relationship.
+#' One of `dataset_1`, `dataset_2` or `none`.
```
gogonzo, line 20: I think it could be more precise (even though I don't like the implementation)

```suggestion
#' @param parent (`character(1)`) Indicates which argument hold a name of a parent dataset in a parent-child
#' relationship or `none` if it is an undirected relationship.
#' One of `"dataset_1"`, `"dataset_2"` or `"none"`.
```

R/join_key.R
```diff
@@ -2,30 +2,35 @@
...
 #'
 #' @export
 #'
 #' @examples
 #' join_key("d1", "d2", c("A"))
 #' join_key("d1", "d2", c("A" = "B"))
 #' join_key("d1", "d2", c("A" = "B", "C"))
-join_key <- function(dataset_1, dataset_2 = dataset_1, keys) {
+join_key <- function(dataset_1, dataset_2 = dataset_1, keys, parent = c("dataset_1", "dataset_2", "none")) {
+  parent <- match.arg(parent)
```
gogonzo, line 33: I find it little odd with proposed way of specifying parent. `parent` is not exactly a parent but it rather sets the direction of the relationship between `dataset_1` and `dataset_2`. Here, information in `parent` is redundant and could be easily replaced by inverting `dataset_1` with `dataset_2`. I think there is too many (two :D) ways to achieve the same result. I propose

```suggestion
join_key <- function(dataset_1, dataset_2 = dataset_1, keys, directed = TRUE) {
  checkmate::assert_flag(directed)
```

And instead of doing
```r
# setting B as parent
join_key(dataset1 = "A", dataset_2 = "B", parent = "dataset_2")
join_key(dataset1 = "B", dataset_2 = "A") 

# setting no parent
join_key(dataset1 = "A", dataset_2 = "B", parent = "none")
```

I would prefer only

```r
# setting B as parent 
join_key(dataset1 = "B", dataset_2 = "A")

# setting no parent
join_key(dataset1 = "A", dataset_2 = "B", directed = FALSE)
```

R/join_keys-extract.R
```diff
@@ -134,12 +138,13 @@
...
 #'
-#' jk["ds4", "ds1"] <- c("pk4" = "pk1")
+#' jk["ds1", "ds4"] <- c("pk1" = "pk4")
 #'
 #' # Removing a key ---
 #'
 #' jk["ds5", "ds5"] <- NULL
-`[<-.join_keys` <- function(x, i, j, value) {
+`[<-.join_keys` <- function(x, i, j, parent = c("i", "j", "none"), value) {
+  parent <- match.arg(parent)
```
gogonzo, line 147: `parent` have different default values and thus explaination of the same argument differs between `[<-.join_keys` and `join_key`

```suggestion
`[<-.join_keys` <- function(x, i, j, directed = TRUE, value) {
  checkmate::assert_flag(directed)
```

R/join_keys-extract.R
```diff
@@ -163,8 +168,19 @@
...
+  if (is.null(value)) {
+    x[[i]][[j]] <- NULL
+    return(x)
+  }
+
+  parent_conversion <- switch(parent,
+    i = "dataset_1",
+    j = "dataset_2",
+    "none"
+  )
```
gogonzo, line 181: For obvious reason, nobody likes this line. `directed` would be cleaner

gogonzo (changes requested):
`parent` argument is confusing. Default values differ between two functions and the possibility of having a parent as `dataset_1` or `dataset_2` is too "freely". I would prefer to limit `dataset_1` to be a parent only. So that

```r
join_key("parent", "child", c(id_in_parent = id_in_child), directed = TRUE) # default

join_key("A", "B", c(id_in_A = id_in_B), directed = FALSE)
```

R/join_keys-c.R
```diff
@@ -10,7 +10,7 @@
 #'   jk,
 #'   join_keys(
 #'     join_key("ds4", keys = c("pk4", "pk4_2")),
-#'     join_key("ds4", "ds3", c(pk4_2 = "pk3"))
+#'     join_key("ds3", "ds4", c(pk3 = "pk4_2"))
 #'   )
 #' )
 c.join_keys <- function(...) {
```
chlebowa, line 16: Is there a reason keys are concatenated `first < add (empty < add second < add third < ...)`? I naively think all of `...` can be put into the `Reduce` but I suppose I'm missing something.

R/join_keys-extract.R
chlebowa: This is a solid argument. The `match.arg` pattern was my preferred solution but I'm not convinced anymore.

R/join_key.R
averissimo: I like the explicit usage of indicating if it's one or the other, but given that this will hardly ever be used I don't oppose the flag. Especially as this avoids the issue of different parameters values.

I'll make the changes, which I believe will resolve all the comments from your review.

R/join_key.R
```diff
@@ -2,30 +2,35 @@
 #'
 #' @description `r lifecycle::badge("stable")`
 #'
-#' @details `join_key()` will create a relationship for the variables on a pair
-#' of datasets.
+#' @description `Create a relationship for the variables on a pair of datasets.
+#' It is directed by default with `dataset_1` being the parent.
```
kartikeyakirar, line 6:
```suggestion
#' @description Create a relationship between two datasets, dataset_1 and dataset_2.
#' By default, this function establishes a directed relationship with dataset_1 as the parent.
#' If dataset_2 is not specified, the function creates a primary key for dataset_1.
```

R/join_keys-extract.R
averissimo: Pushed new version with `directed = TRUE` parameter.

R/join_key.R
```diff
@@ -2,33 +2,40 @@
...
 #' @examples
 #' join_key("d1", "d2", c("A"))
 #' join_key("d1", "d2", c("A" = "B"))
 #' join_key("d1", "d2", c("A" = "B", "C"))
-join_key <- function(dataset_1, dataset_2 = dataset_1, keys) {
+join_key <- function(dataset_1, dataset_2 = dataset_1, keys, directed = TRUE) {
   checkmate::assert_string(dataset_1)
   checkmate::assert_string(dataset_2)
   checkmate::assert_character(keys, any.missing = FALSE)
+  checkmate::assert_logical(directed)
```
kartikeyakirar, line 38:
```suggestion
  checkmate::assert_flag(directed)
```

R/join_keys-c.R
kartikeyakirar, line 16: @averissimo Here's a shorter version for c.join_keys. WDYT?

```
c.join_keys <- function(...) {
  args <- rlang::list2(...)  #Combined list2 calls for different subsets to reduce overhead.
  checkmate::assert_multi_class(args[[1]], classes = c("join_keys", "join_key_set"))
# Removed assert_list check for additional arguments as they're processed in Reduce function.

# Using the reduce function on the args list avoids unnecessary steps of separating and merging the first argument, simplifying the process.

  out <- Reduce(function(.x, .y) {
    # or use checkmate::assert_list()
    checkmate::assert_multi_class(.y, classes = c("join_keys", "join_key_set"))

    # Perform the merge operation
    merged <- utils::modifyList(.x, .y, keep.null = FALSE)
    parents(merged) <- .merge_parents(.x, .y)
    merged
  }, args, init = join_keys())

  out
}
```

R/join_keys-extract.R
```diff
@@ -134,12 +138,13 @@
...
 #'
-#' jk["ds4", "ds1"] <- c("pk4" = "pk1")
+#' jk["ds1", "ds4"] <- c("pk1" = "pk4")
 #'
 #' # Removing a key ---
 #'
 #' jk["ds5", "ds5"] <- NULL
-`[<-.join_keys` <- function(x, i, j, value) {
+`[<-.join_keys` <- function(x, i, j, directed = TRUE, value) {
+  checkmate::assert_logical(directed)
```
kartikeyakirar, line 147:
```suggestion
  checkmate::assert_flag(directed)
```

R/join_keys-utils.R
```diff
@@ -111,44 +111,31 @@ update_keys_given_parents <- function(x) {
 
   datanames <- names(jk)
   duplicate_pairs <- list()
```
kartikeyakirar, line 113: could we allocate the size here instead of increasing it loop something like this
```
len <- length(datanames)
  duplicate_pairs <- vector("list", length = len * (len - 1) / 2)
  d_ix <- 1
```

and instead of 
`duplicate_pairs <- append(duplicate_pairs, paste(d1, d2))`
do this
```
duplicate_pairs[[d_ix]] <-  paste(d1, d2)
d_ix <- d_ix+1
```

R/join_key.R
```diff
@@ -65,6 +72,12 @@ join_key <- function(dataset_1, dataset_2 = dataset_1, keys) {
     keys <- NULL
   }
 
+  parents <- if (directed && dataset_1 != dataset_2) {
+    structure(list(dataset_1), names = dataset_2)
```
gogonzo, line 76: `setNames` is more precise
```suggestion
    setNames(list(dataset_1), dataset_2)
```

R/join_keys-parents.R
```diff
@@ -73,11 +73,15 @@ parents.teal_data <- function(x) {
 #' parents(jk)["ds6"] <- "ds5"
 #' parents(jk)["ds7"] <- "ds6"
 `parents<-.join_keys` <- function(x, value) {
-  checkmate::assert_list(value, types = "character", names = "named")
+  checkmate::assert_list(value, types = c("character"), names = "named")
```
gogonzo, line 76: I think this extra call is unnecessary. What's special about adding `c()`?

```suggestion
  checkmate::assert_list(value, types = "character", names = "named")
```

R/join_keys-c.R
averissimo: The reason why we are breaking the first argument from the rest is refered in the [comment above](https://github.com/insightsengineering/teal.data/pull/215/files#r1426659206).

_To sum it up_: we allow a unrestrictive merge between argument 1 and "the rest". The rest must not have any conflicting keys between themselves.

Short example of this in play can be read in the [same comment above](https://github.com/insightsengineering/teal.data/pull/215/files#r1426659206)

🔴 If this behavior is not necessary, them we can simplify it.

averissimo: @insightsengineering/nest-core-dev any thoughts on whether we should change this behavior?

- as is (restrictive)
- unrestrictive for `c(...)` and `join_keys(...)`
- unrestrictive for `c(..)` call only

averissimo: note: the assertions are all in the top-level function to allow for more informative error messages, otherwise it would show `Error in f(init, x[[i]])` as it's define inside the Reduce function

Toy example:

``` r
and <- function(...) {
  data <- rlang::list2(...) |> purrr::flatten() |> unlist()
  checkmate::assert_logical(data)
  
  Reduce(
    f = \(.x, .y) {
      checkmate::assert_multi_class(.y, classes = c("join_keys", "join_key_set"))
      .x && .y
    },
    x = data,
    init = TRUE
  )
}

and(TRUE, 1)
#> Error in and(TRUE, 1):
#>   Assertion on 'data' failed: Must be of type 'logical', not 'double'.
and(TRUE, FALSE)
#> Error in f(init, x[[i]]):
#>   Assertion on '.y' failed: Must inherit from class 'join_keys'/'join_key_set', but has class 'logical'.
```

R/join_keys-utils.R
averissimo: This function would benefit from some refactoring, but I don't think we get any performance benefits with the expected size of `join_keys`.

Given no performance boost _(assuming benchmarks below are realistic)_, I find it easier to read the code as is. I'm open to discuss it though :+1: 

The individual operations are more efficient as `f{1,2}_manual()` shows, however, that boost is completely lost when accounting for the initialization of the vector.

At least not until we get into real big vectors (see second benchmark), and still it needs to go up to absurd sizes.

``` r
ntimes <- 100L

f1 <- local({
  l <- NULL
  \() {
    if (is.null(l)) {
      print("initializing f1")
      l <<- list()
    }
    l <<- append(l, paste("ad", "bc"))
    invisible(l) # returning this or NULL does not affect benchmark
  }
})

f2 <- local({
  l <- NULL
  d_ix <- 1
  \() {
    if (is.null(l)) {
      print("initializing f2")
      l <<- vector("list", length = ntimes)
    }
    l[[d_ix]] <- paste("ad", "bc")
    d_ix <<- d_ix + 1
    invisible(l) # returning this or NULL does not affect benchmark
  }
})

f1_manual_l <- rep(paste("ad", "bc"), ntimes - 1) |> as.list()
f2_manual_l <- vector("list", length = ntimes)
d_ix <- length(f2_manual_l)

f1_manual <- \() invisible(append(f1_manual_l, paste("ad", "bc")))
f2_manual <- \() {
  f2_manual_l[[d_ix]] <- paste("ad", "bc")
  d_ix <- d_ix + 1
  invisible(f2_manual_l)
}

microbenchmark::microbenchmark(f1(), f2(), f1_manual(), f2_manual(), times = ntimes)
#> [1] "initializing f2"
#> [1] "initializing f1"
#> Unit: microseconds
#>         expr   min     lq     mean median     uq      max neval
#>         f1() 2.067 2.2405  2.58464 2.3640 2.4955   15.255   100
#>         f2() 2.181 2.2600  3.22357 2.2985 2.3720   87.602   100
#>  f1_manual() 2.285 2.3765 91.59518 2.4225 2.5110 8911.203   100
#>  f2_manual() 1.723 1.7850 22.44896 1.8250 1.8725 2055.476   100
```

``` r
ntimes <- 10000L

# ...

microbenchmark::microbenchmark(f1(), f2(), f1_manual(), f2_manual(), times = ntimes)
#> [1] "initializing f1"
#> [1] "initializing f2"
#> Unit: microseconds
#>         expr    min      lq     mean  median      uq      max neval
#>         f1()  2.186 16.0270 31.30275 29.4425 43.3985 1861.159 10000
#>         f2() 42.127 44.5890 52.01254 45.8570 50.1740 4813.170 10000
#>  f1_manual() 46.798 48.8735 56.40130 50.1150 55.0195 8137.664 10000
#>  f2_manual() 41.856 44.0300 51.30092 45.2985 49.3840 3963.022 10000
```

R/join_key.R
```diff
@@ -2,33 +2,40 @@
...
+#' a primary key for `dataset_1` is created.
+#' @param keys (optionally named `character`) Column mapping between the datasets,
+#' where `names(keys)` maps columns in `dataset_1` corresponding to columns of
+#' `dataset_2` given by the elements of `keys`.
 #'
 #' If unnamed, the same column names are used for both datasets.
 #'
 #' If any element of the `keys` vector is empty with a non-empty name, then the name is
 #' used for both datasets.
+#' @param directed (`logical(1)`) Flag that indicates whether it should create
```
chlebowa, line 19: I'm not sure `directed` is the best possible name.
Could we consider some alternatives?
`directional`
`hierarchical`
`parental`
`vertical`

R/join_key.R
```diff
@@ -65,6 +72,12 @@ join_key <- function(dataset_1, dataset_2 = dataset_1, keys) {
     keys <- NULL
   }
 
+  parents <- if (directed && dataset_1 != dataset_2) {
+    setNames(list(dataset_1), dataset_2)
```
chlebowa, line 76:
```suggestion
    stats::setNames(list(dataset_1), dataset_2)
```

R/join_key.R
```diff
@@ -65,6 +72,12 @@ join_key <- function(dataset_1, dataset_2 = dataset_1, keys) {
...
+  } else {
+    list()
+  }
+
   structure(
     list(
       structure(
```
chlebowa, line 83: If we are going back to `setNames`, I'm afraid you have to change all instances.

R/join_keys-parents.R
```diff
@@ -78,6 +78,10 @@ parents.teal_data <- function(x) {
   new_parents <- list()
 
   for (dataset in names(value)) {
+    if (checkmate::test_scalar_na(value[[dataset]])) {
+      checkmate::assert("May not contain `NA_character_`", .var.name = "value")
```
chlebowa, line 82:
```suggestion
      checkmate::assert("Must not contain `NA_character_`", .var.name = "value")
```

R/join_keys-parents.R
```diff
@@ -78,6 +78,10 @@ parents.teal_data <- function(x) {
   new_parents <- list()
 
   for (dataset in names(value)) {
+    if (checkmate::test_scalar_na(value[[dataset]])) {
```
chlebowa, line 81: It seems this could be covered by `assert_string`. Is it because the custom error message?

vignettes/join-keys.Rmd
```diff
@@ -212,22 +227,19 @@ join_keys(td_fk) <- join_keys(
...
 
 ### Implicit Relationships
 
-Two datasets that share common _foreign_ keys to the same _parent_ dataset have an implicit relationship between them that is modeled and accessible in `join_keys`.
+Two datasets that share common _foreign_ keys to the same _parent_ dataset have an implicit relationship between them that is modeled and accessible in _joining keys_.
 
-This is a special inferred relationship from existing `join_keys` that does not need to be explicitly defined and can be seamlessly accessible just as any other foreign key.
-As any other foreign key it can be overwritten.
+This is a special inferred relationship from existing foreign keys that does not need to be explicitly defined and can be seamlessly accessible just as any other foreign key.
+As any other foreign key they can be overwritten.
```
chlebowa, line 242:
```suggestion
This is a special relationship that is inferred from existing foreign keys.
It does not need to be explicitly defined but it can be accessed and overwritten just as any other foreign key.
```

R/join_keys.R
```diff
@@ -6,13 +6,15 @@
 #' join_keys(...)
 #'
 #' @description
-#' `join_keys()` facilitates the creation and retrieval of relationships between datasets.
-#' `join_keys` class extends a list and contains keys connecting pairs of datasets. Each element
-#' of the list contains keys for specific dataset. Each dataset can have a relationship with
-#' itself (primary key) and with other datasets.
+#' Facilitates the creation and retrieval of relationships between datasets.
+#' `join_keys` class extends a list and contains keys connecting pairs of datasets.
```
chlebowa, line 10:
```suggestion
#' `join_keys` class extends `list` and contains keys connecting pairs of datasets.
```

R/join_keys-c.R
kartikeyakirar: >  If this behavior is not necessary, them we can simplify it.

Concatenated operation overriding the relationship when there are two datasets e.g in  `case1` and `case2,` but it throws an error in `case3` which is a bit odd. This may create confusion, so simplifying it and not restricting this behavior unlike join_keys() constructor might be better approach IMO.

Another argument for this is that we should not put a safeguard for everything, and this might be an intentional operation from the app-dev side.

```
##case1:################################
c(
  join_keys(join_key("ds1", "ds1", "original")),
  join_keys(
    join_key("ds1", "ds1", "first_change")
  )
) |> invisible()

##case2:################################
c(c(
  join_keys(join_key("ds1", "ds1", "original")),
  join_keys(
    join_key("ds1", "ds1", "first_change")
  )),
  join_keys(
    join_key("ds1", "ds1", "second change")
  )
) 

##case3:################################
c(
  join_keys(join_key("ds1", "ds1", "original")),
  join_keys(
    join_key("ds1", "ds1", "first_change")
  ),
  join_keys(
    join_key("ds1", "ds1", "second change")
  )
) |> invisible()
#> Error in stop_message(dataset_1_one, dataset_2_one):
#>   cannot specify multiple different join keys between datasets: ds1 and ds1
```

kartikeyakirar: I think both `c(...)` and `join_keys(...)` should be unrestricted. or atleast for `c(...)` as its confusing to see error for more then two datasets as [described here](https://github.com/insightsengineering/teal.data/pull/215#discussion_r1432806727).

kartikeyakirar: > note: the assertions are all in the top-level function to allow for more informative error messages, otherwise it would show Error in f(init, x[[i]]) as it's define inside the Reduce function

we can add meaning full message in assertion there

R/join_keys-utils.R
kartikeyakirar: > This function would benefit from some refactoring, but I don't think we get any performance benefits with the expected size of join_keys.

Yeah, I did not consider the size, so I don't think the list of join_keys will be very long to make any significant difference.

averissimo: That variable was not actually being used :facepalm: 

Regardless, went a bit deeper and tested a real benchmark actually filling the list at each run (and without cold starts)

I'm so sure anymore about the benefit (even for small sizes)

``` r
b <- \(max_size = 10, ntimes = 100L) {
  f1 <- local({
    l <- NULL
    l_call <- \(.x) {
      if (is.null(l)) l <<- list()
      l <<- append(l, paste("ad", "bc"))
      TRUE
    }
    \() vapply(seq_len(max_size), l_call, logical(1))
  })
  
  f2 <- local({
    l <- NULL
    d_ix <- 1
    l_call <- \(.x) {
      if (is.null(l)) l <<- vector("list", length = max_size)
      l[[d_ix]] <<- paste("ad", "bc")
      d_ix <<- d_ix + 1
      TRUE
    }
    \() vapply(seq_len(max_size), l_call, logical(1))
  })
  
  microbenchmark::microbenchmark(
    f1(), 
    f2(), 
    times = ntimes, 
    setup = {
      list()
      vector("list", length = max_size)
    }
  )
}

range <- 1:100

res <- Reduce(
  \(x, y) {
    res <- summary(b(y))
    dplyr::bind_rows(x, data.frame(max_size = y, f1 = res$mean[[1]], f2 = res$mean[[2]]))
  },
  x = range,
  init = data.frame(max_size = numeric(0L), f1 = numeric(0L), f2 = numeric(0L))
)

library(ggplot2)
res |> 
  dplyr::mutate(ratio = f1 / f2) |> 
  dplyr::rename(append = f1, assign = f2) |> 
  tidyr::pivot_longer(cols = c(append, assign)) |> 
  dplyr::mutate(
    ratio = dplyr::if_else(name == "append", as.character(round(ratio, digits = 2)), NA_character_),
    ratio = dplyr::if_else(max_size %% 5 == 0, ratio, NA_character_)
  ) |> 
  
  ggplot(mapping = aes(x = max_size, y = value, color = name, label = ratio)) + 
    geom_point() +
    ggrepel::geom_label_repel(min.segment.length = 0) + 
    scale_y_continuous(trans = "log10", name = "Mean in microseconds (log10 scale)") +
    labs(subtitle = "label is ratio between append / assign") +
    theme_minimal()
#> Warning: Removed 180 rows containing missing values (`geom_label_repel()`).
```



<sup>Created on 2023-12-20 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

R/join_key.R
averissimo: I thought about that before accepting the suggestion from @gogonzo, however this doesn't mix the 2 on the _same_ expression.

`setNames` seems fine for such a simple variable construct. 

If you feel strongly about this, I'd prefer reverting to `structure` above.

R/join_keys-parents.R
averissimo: Yes, but now that I look at this we can tweak with the `.var.name`

I'm changing to the assert below

``` r
dataset <-"d1"
value <- list(d1 = NA_character_)
checkmate::assert_string(value[[dataset]], .var.name = sprintf("value[[\"%s\"]]", dataset))
#> Error in eval(expr, envir, enclos): Assertion on 'value[["d1"]]' failed: May not be NA.
```

R/join_key.R
averissimo: I think `directed` is a good name, but I'm not married. Maybe we can have a quick poll with emojis

React to this comment with:

:+1:  _for_ `directed`
:smile:  _for_ `directional`
:tada:  _for_ `hierarchical`
:heart:  _for_ `parental`
:rocket:  _for_ `vertical`
:eyes: _for_ `dataset_1_as_parent`

gogonzo: "directed" is a legit term in a graph theory which join_keys follows. We also use other graph theory terms, like cyclic or topological sort. Directed is also short enough unlike "directional" or other proposed

R/join_keys-parents.R
chlebowa: Looks great :+1:

R/join_key.R
chlebowa: If it's a technical term (which I was unaware of), let's keep it, although I would question whether it is widely known.

R/join_key.R
chlebowa: You're right. Sorry, I was reading too fast.

R/join_keys-c.R
gogonzo: Now parents is always a list (empty or not, but never NULL), so I believe you can remove `.merge_parents` and replace it with `modifyList(.x, .y)`

R/join_keys-extract.R
```diff
@@ -234,9 +244,14 @@
   # Remove classes to use list-based get/assign operations
   new_x <- unclass(x)
 
-  # In case a pair is removed, also remove the symmetric pair
+  # In case a pair is removed, also remove the symmetric pair and update parents
   removed_names <- setdiff(names(new_x[[i]]), names(norm_value))
-  for (.x in removed_names) new_x[[.x]][[i]] <- NULL
+  for (.x in removed_names) {
+    if (identical(parent(x, .x), i)) attr(new_x, "parents")[[.x]] <- NULL
+    if (identical(parent(x, i), .x)) attr(new_x, "parents")[[i]] <- NULL
```
gogonzo, line 251:
```suggestion
    if (identical(parent(x, .x), i)) parents(new_x)[[.x]] <- NULL
    if (identical(parent(x, i), .x)) parents(new_x)[[i]] <- NULL
```

tests/testthat/test-join_keys-c.R
```diff
@@ -118,17 +118,16 @@ testthat::test_that("c.join_key_set merges with empty and non-empty parents", {
...
 
   expected <- join_keys(
     join_key("d1", "d1", "a"),
     join_key("d3", "d3", "c"),
     join_key("d4", "d4", "d"),
-    join_key("d3", "d4", "cd")
+    join_key("d3", "d4", "cd", directed = FALSE)
   )
-  parents(expected) <- list(d3 = "d4")
+  parents(expected) <- list(d4 = "d3")
```
gogonzo, line 130: test for expect equal is reduntant. When c(jk1, jk2) is identical to expected it is certain that next testthat is true.

tests/testthat/test-join_keys-extract.R
```diff
@@ -373,21 +393,19 @@ testthat::test_that("[[<-.join_keys removes keys with NULL", {
...
-    join_key("d2", "d1", "A"),
-    join_key("d2", "d3", "B"),
-    join_key("d3", "d2", "B")
+    join_key("d2", "d3", "B")
   )
   my_keys[["d1"]][["d2"]] <- NULL
 
-  testthat::expect_identical(
+  expect_null(my_keys["d1", "d2"])
+  expect_null(my_keys["d2", "d1"])
```
gogonzo, line 404: This is redundant when expect_equal is true.
```suggestion
```

tests/testthat/test-join_keys-c.R
averissimo: It is. The test expanded the scope to test if both merges are equal. Removed!

R/join_keys-c.R
averissimo: I'll make the proposed change @kartikeyakirar , that's a good argument :+1:

averissimo: Please check b3177082

R/join_keys-c.R
averissimo: Yup. done!

kartikeyakirar:
@averissimo PR looks great. I have re-reviewed it and tested it with `c(...) `and `join_keys(...)`. Additionally, I found three unused variables (by running the linter), which I have removed here [7947878](https://github.com/insightsengineering/teal.data/pull/215/commits/79478789f9b30e2669c8866f0bbdda8e16dad3c8)

kartikeyakirar (approved):
👍

PR #219, opened 2023-12-12
bug fix: `teal_data` appears empty when app starts 

kartikeyakirar:
this is fixes: https://github.com/insightsengineering/teal/issues/985

**Problem Description:**
When launching the app or initially loading a module, `get_code_dependency `  triggers console warnings. This issue is  the warning message:
```
Warning in teal.data:::get_code_dependency(attr(datasets, "preprocessing_code"),  :
  Objects not found in 'qenv' environment: IRIS, NPK

```
The problem occurs during the execution of the following code:
```
app <- init(
  data = teal_data(IRIS = iris, NPK = npk),
  modules = modules(
    example_module(label = "example 1"),
    example_module(label = "example 2")
  )
)

runApp(app)
```
This issue arises in the `teal_data_to_filtered_data` function, which extracts code to piggyback the entire preprocessing sequence. However, if `teal_data` is initialized without embedded code, it lacks the preprocessing code. This can be observed in the screenshot below:
image

**Resolution:**
Initializing teal_data with code resolves this issue, as shown below:
```
data <- teal_data(IRIS = iris,
                  MTCARS =mtcars,code = "IRIS = iris\nMTCARS =mtcars")
```

**Proposed Solution:**
I modified the `teal_data `constructor to automatically extract the '...' (ellipsis) argument and convert it into a code string, which is then set in the code slot. This approach resolves the issue. However, I am not able to find solution to directly in `teal`, as the ellipsis argument cannot be accessed there and is only valid in the constructor.

This fix seems to resolve the issue without introducing new problems. Let me know your thoughts and feedback on this solution.

### DIFF
--- a/R/teal_data.R
+++ b/R/teal_data.R
@@ -49,6 +49,15 @@ teal_data <- function(...,
     if (length(data_objects) > 0 && !checkmate::test_names(names(data_objects), type = "named")) {
       stop("Dot (`...`) arguments on `teal_data()` must be named.")
     }
+
+    if(length(data_objects) > 0 && length(code) == 0) {
+      arg_list <- match.call(expand.dots = FALSE)$`...`
+      code_string <- sapply(names(data_objects), function(name) {
+        argument_string <- arg_list[[name]]
+        paste0(name, " <- ", argument_string)
+      }, USE.NAMES = FALSE)
+      code <- paste(code_string, collapse = "\n")
+    }
     new_teal_data(
       data = data_objects,
       code = code,
--- a/tests/testthat/test-get_code.R
+++ b/tests/testthat/test-get_code.R
@@ -3,7 +3,7 @@ warning_message <- "warning('Code was not verified for reproducibility.')"
 testthat::test_that("get_code with datanames handles empty @code slot", {
   testthat::expect_identical(
     get_code(teal_data(a = 1, code = character(0)), datanames = "a"),
-    warning_message
+    paste(warning_message, "a <- 1", sep  = "\n")
   )
   testthat::expect_identical(
     get_code(teal_data(a = 1, code = ""), datanames = "a"),


kartikeyakirar:
**Test results:**

```
#################################################
# without code
#################################################
data <- teal_data(IRIS = iris,
                  MTCARS =mtcars)
> data@code
[1] "IRIS <- iris\nMTCARS <- mtcars"

#################################################
# with code
#################################################
data <- teal_data(IRIS = iris,
                  MTCARS =mtcars,code = "IRIS <- iris\nMTCARS <- mtcars")
> data@code
[1] "IRIS <- iris\nMTCARS <- mtcars"


#################################################
# using within 
#################################################
data <- teal_data()
data <- within(data, {
  IRIS <- iris
  MTCARS <- mtcars
})
> data@code
[1] "IRIS <- iris\nMTCARS <- mtcars"
```

m7pr:
Looks like a quick and proper fix. The only thing I am afraid of is whether we want a situation where the code is appended for us and probably in most cases it's gonna be redundant like `teal_data(ADSL = ADSL)` and then the code is `ADSL = ADSL`. I think we should somehow fix `get_code_dependency` so it does not throw this error on startup - I think it lacks of `datanames` attribute in which it seeks for valid dataname objects. So we either extend `datanames` creation automatically for the user, or we force a user to specify `datanames` before `teal_data` is used in app. What if `teal::init` had a check for `teal_data` not having any `datanames`?

m7pr:
Maybe this will actually fix the issue? https://github.com/insightsengineering/teal/pull/1004

chlebowa (changes requested):
This is no solution at all.
1) The user is allowed to pass datasets and code or evaluate code to create datasets. Code must not be derived for them.
2) As you yourself admit, it does not cover the case of a `teal_data` created empty.
3) This will fail if the dataset is an object that not exist on the search path (literals don't count).

```
devtools::load_all("../teal.data")
M <- mtcars
td <- teal_data(i = iris, m = M)
rm(M)
get_code(td) %>% cat(sep = "\n")
get_code(td, deparse = FALSE) %>% eval

Error in eval(.) : object 'M' not found
In addition: Warning message:
In eval(.) : Code was not verified for reproducibility.
```

m7pr:
I would opt to wait what would be the outcome of https://github.com/insightsengineering/teal/pull/1004

chlebowa:
> Maybe this will actually fix the issue? [insightsengineering/teal#1004](https://github.com/insightsengineering/teal/pull/1004)
> I would opt to wait what would be the outcome of [insightsengineering/teal#1004](https://github.com/insightsengineering/teal/pull/1004)

No it won't. These issues are not connected at all.

kartikeyakirar:
> The only thing I am afraid of is whether we want a situation where the code is appended for us and probably in most cases it's gonna be redundant like teal_data(ADSL = ADSL) and then the code is ADSL = ADSL. 

I was thinking that if a developer does not want to add code for some reason, they can specify `code = "" `and it won't automatically get code from the ellipsis arg. (I'm a bit skeptical about keeping this as an option, but I thought that if someone specifies the code argument, they are responsible and aware.)

> I think we should somehow fix get_code_dependency so it does not throw this error on startup - I think it lacks of datanames attribute in which it seeks for valid dataname objects. So we either extend datanames creation automatically for the user, or we force a user to specify datanames before teal_data is used in app. What if teal::init had a check for teal_data not having any datanames?

PR https://github.com/insightsengineering/teal/pull/1004 already solved this issue , it automatically sets the dataname from data@env if it is not found and performs a check. This part of the request has already been completed. The `get_code_dependency` throws a warning, but it is not the cause of the issue. When we create `teal_data` with  `within`, we always have the code available. However, if we create `teal_data` directly, we did not have a mechanism to add code automatically if it is not specified. This pull request solves that issue.

kartikeyakirar:
> The user is allowed to pass datasets and code or evaluate code to create datasets. Code must not be derived for them.This will fail if the dataset is an object that not exist on the search path (literals don't count).

@chlebowa  In this particular scenario, an alternative solution would be to avoid automatically deriving the code from the argument altogether. Instead, in the` teal `package when `get_code_dependency`  is called https://github.com/insightsengineering/teal/blob/982184d76cd002156d264f5219ad07837136a7fc/R/get_rcode_utils.R#L49, I will check if the `preprocessing_code` attribute contains any code. If it doesn't, then `get_code_dependency` won't be executed and there will be no warnings. WDYT?

Otherwise I'm curious if you have any suggestions to make this approach work better?

chlebowa:
In my opinion this is a poor approach to begin with and there is no use trying to make it work.

m7pr:
So if this is caused by 

```r
teal.data:::get_code_dependency(attr(datasets, "preprocessing_code"), names = datanames)
```

and `attr(datasets, "preprocessing_code")` can be NULL then we should fix `teal.data:::get_code_dependency` for NULL input, or set  `attr(datasets, "preprocessing_code")` to be `character(0)` is this is empty or `""` as `teal.data:::get_code_dependency` can handle that!

```r
> teal.data:::get_code_dependency("", 'a')
[1] ""
> teal.data:::get_code_dependency(character(0), 'a')
character(0)
> teal.data:::get_code_dependency(NULL, 'a')
Error in teal.data:::get_code_dependency(NULL, "a") : 
  Assertion on 'code' failed: Must be of type 'character', not 'NULL'.
```

kartikeyakirar:
> and attr(datasets, "preprocessing_code") can be NULL then we

 attr(datasets, "preprocessing_code") is wither "" or "warning('Code was not verified for reproducibility.')" based on varified slot

chlebowa:
Actually, it can only be `""` if there is no data.

m7pr:
Ok @kartikeyakirar but this is only a warning, and the result of `get_code_dependency` is `character(0)`

chlebowa:
> Ok @kartikeyakirar but this is only a warning, and the result of `get_code_dependency` is `character(0)`

What's your point? Removing the warning is the issue.

m7pr:
I though it brings errors not warnings. We can remove the warning from `get_code_dependency` - it is an internal function not exposed to the user so will be used in a controlled way, so maybe there is no need to have a warning inside it?

kartikeyakirar:
> We can remove the warning from get_code_dependency - it is an internal function not exposed to the user so will be used in a controlled way,

The function `get_code_dependency` is used in two places: once in the `get_code` method for `teal_data` and once in `get_datasets_code`. In some cases, if there is code but it does not match with datanames, it may throw a warning twice. As this function is internal, I support for removing warning.

m7pr:
We can also consider

- only presenting this warning in an interactive usage (non-shiny usage) with `shiny::isRunning` verification but that brings a dependency on `shiny`
- parametrizing if we want to do this check with `check=TRUE` parameter, set to `FALSE` in usage within teal https://github.com/insightsengineering/teal.data/pull/223

chlebowa:
> We can also consider
> 
> * only presenting this warning in an interactive usage (non-shiny usage) with `shiny::isRunning` verification but that brings a dependency on `shiny`
> * parametrizing if we want to do this check with `check=TRUE` parameter, set to `FALSE` in usage within teal [985 parametrize name check in get_code_dependency #223](https://github.com/insightsengineering/teal.data/pull/223)

This is the best solution so far IMO :ok_hand:

m7pr:
So far provided this `check_names` parameter in here https://github.com/insightsengineering/teal.data/pull/223 and in teal https://github.com/insightsengineering/teal/pull/1012

m7pr:
@kartikeyakirar are you OK closing this one, and we will merge the other two that I mentioned in the previous message?

kartikeyakirar:
@kartikeyakirar are you OK closing this one, and we will merge the other two that I mentioned in the previous message?

Yup! The solution looks good

PR #223, opened 2023-12-15
985 parametrize name check in get_code_dependency

m7pr:
Alternative for https://github.com/insightsengineering/teal.data/pull/219 potentially fixes https://github.com/insightsengineering/teal/issues/985 if we extend `get_code_dependency` usage with `check_names = FALSE` in teal - provided in https://github.com/insightsengineering/teal/pull/1012


chlebowa (approved):
:+1:

PR #230, opened 2023-12-18
220 detect default side effects

### LINKED ISSUE #220: [BUG] get_code should return side-effect line always.
```r
data <- teal_data() |>
  within({
    set.seed(1)
    library(scda)
    library(airway)
    library(MultiAssayExperiment)
    data(miniACC, envir = environment())
    data(airway, envir = environment())
    any_variable <- "elo"
    ADSL <- synthetic_cdisc_data("latest")$adsl
    ADSL$categorical <- sample(letters[1:3], size = nrow(ADSL), replace = TRUE, prob = c(.1, .3, .6))
    ADTTE <- synthetic_cdisc_data("latest")$adtte
    ADRS <- synthetic_cdisc_data("latest")$adrs
  })

datanames(data) <- c("ADSL", "ADTTE", "airway", "ADRS", "miniACC", "any_variable")

get_code(data, datanames = "airway")
get_code(data, datanames = "ADRS")
```

Such calls like `library(...)`, `data(...)` without `# linkto` tag should be always returned when extracting code. At this moment they are never returned if they have no "tag".

m7pr:
Close #220

We introduced a list of functions that are treated as default side effects having impact on all objects. Those will be always returned with `get_code`. Currently the list contains of `library`, `require` and `data`. I wonder if this should be a parameter to `get_code` so that this list can be extended and adjusted by future users?


m7pr:
Yeah, had similar thoughts. I think the most flexible will be an exposure
of a parameter vector in get_code that would take a list of names that are
always returned. This could default to 'library' and be changed by a user.

chlebowa:
Let's try to think of something else before we add arguments for everything.

m7pr:
It's always the matter of what are the cases and needs of users. I think in
this specific situation there are a lot of unforeseen situations in which a
user would assume some specific function calls should be returned because
they are needed. I think libraries and require are fine for a good start.
Then maybe `options` and `set.seed`. But what else? There might be a lot
and possibly each user has different needs. That's why I am proposing a
parameter. This function covers a very complicated problem so I would
imagine we can have 4 or 5 parameters. Simple `lm` function for linear
regression has 12 parameters.

I don't think we're gonna cover every scenario, even if we think of
something else. I think introduction of parameters is fine if you see such
a situation where the usage has a broad scope.

chlebowa:
If a person has to manage what function calls are "automatically" included, they might as well manually tag them.

A reasonable exception is the search path, that one has to be reproduced.

R/utils-get_code_dependency.R
```diff
@@ -49,9 +49,9 @@ get_code_dependency <- function(code, names, check_names = TRUE) {
   graph <- code_graph(calls_pd)
   ind <- unlist(lapply(names, function(x) graph_parser(x, graph)))
 
-  se_ind <- default_side_effects(calls_pd)
+  lib_ind <- detect_libraries(calls_pd)
 
-  as.character(code[unique(c(se_ind, ind))])
+  as.character(code[unique(c(lib_ind, ind))])
```
gogonzo, line 54: I think this operation could be implemented in `teal`. If anyone adds `library` twice to the `qenv` then we shouldn't fix this I guess. If someone combines `library`, `detach` and `library` again, we shouldn't remove duplicate.

Problem of duplicated libraries is something we create in `teal` by adding `library` calls in `get_rcode_header`

m7pr:
Hey, so far I'm only automatically extracting `library` and `require` calls.
I also extended `data()` function detection that is extracted only for respective objects. See a new test that I added.

R/utils-get_code_dependency.R
m7pr: I'm not sure if I understand. 

> I think this operation could be implemented in teal.

You mean `library` calls detection should be handled by functions from https://github.com/insightsengineering/teal/issues/950 ?

> If anyone adds library twice to the qenv then we shouldn't fix this I guess. If someone combines library, detach and library again, we shouldn't remove duplicate.

We can introduce one more step that checks if a `library` is called more than once on the very same package, and if it is, detect how many detaches there were and figure out how many times we need to skip `library()` calls. I imagine someone can call `library(package1)` N times and `detach(package1)` M times. We can handle that, even though I would make a user tag `detach()` with `# @linksto` tags.

gogonzo:
> You mean library calls detection should be handled by functions from https://github.com/insightsengineering/teal/issues/950 ?

Yes but in general I'm looking to resolve `library` calls problem. In `get_rcode_header` we add all attached packages to the code so the problem is not in `teal_data` object. If we add all attached libraries in `get_rcode_header` than adding any `library()` call in `teal_data` is redundant. Also, adding `libarary` calls to `eval_code` doesn't really reproduce session as there are other packages often attached outside of the `eval_code` call. I don't have any strong opinion how it should be done. It is not a problem of this PR so let's not continue a discussion here but rather in insightsengineering/teal#950

R/utils-get_code_dependency.R
m7pr: Yeah, I would also leave it to the user to mess up the code as they please

m7pr:
ALrighty, let's move there for the conversation about library() simplification. And for this feature, do you think we're good to approve and merge?

gogonzo (commented):
Looks good functional-wise. Please apply tests suggestions and add `lifecycle::experimental` label to the `get_call(datanames)`. We are finishing developing this package at this moment. It goes to CRAN soon 💪

tests/testthat/test-get_code.R
```diff
@@ -544,3 +544,86 @@ testthat::test_that("get_code with datanames understands @ usage and do not trea
...
+    paste(
+      warning_message,
+      "library(scda)",
+      "require(dplyr)",
+      "library(MultiAssayExperiment)",
+      "data(\"mtcars\")",
+      "z <- mtcars",
+      sep = "\n"
+    )
+  )
```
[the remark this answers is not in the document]
m7pr: bdaf92362038d33c847d2c11c405fbb865361f8b

m7pr:
Thanks for the review. Badge was added and test were simplified

gogonzo (approved):

Issue #231, opened 2023-12-18
[Bug]: Remove `get_labels` function


gogonzo:
### What happened?

1. `get_labels` and `col_labels` are doing the same thing. Please remove `get_labels` and include it in "deprecated" file.
2. Additionally please relax `stopifnot(is.data.frame(x))` as other type of objects like `DataFrame` can also have `attr(col, "label")`. The best assertion is probably if `!is.null(colnames(x))`.

PR #232, opened 2023-12-20
231 remove get_labels function

kartikeyakirar:
this fixes https://github.com/insightsengineering/teal.data/issues/231


R/formatters_var_labels.R
```diff
@@ -25,7 +25,7 @@
 #' col_labels(x) <- paste("label for", names(iris))
 #' col_labels(x)
 col_labels <- function(x, fill = FALSE) {
-  stopifnot(is.data.frame(x))
+  checkmate::assert_character(colnames(x), null.ok = FALSE, any.missing = FALSE)
```
chlebowa, line 28: Why not `assert_data_frame`?
Also, `null.ok = FALSE` is the default.

kartikeyakirar: > Why not assert_data_frame?

x can be of  class  `data.frame` or `DataFrame`.

gogonzo (approved):
👍

PR #241, opened 2024-01-03
collate and improve docs

chlebowa:
Collates and improves documentation for `col_labels` and `col_relabel`.

Note: these functions could use some improvements but since they are lifted from `formatters`, they have not been modified.


donyunardi (approved):
lgtm

PR #244, opened 2024-01-03
improve label functions

chlebowa:
Rewrote `col_labels`, `col_;labels<-` and `col_relabel`.
All existing unit tests pass.

This is only a suggestion, the functions do their job properly. They have been copied from `formatters`, which is an argument _against_ these changes.

### DIFF
--- a/R/formatters_var_labels.R
+++ b/R/formatters_var_labels.R
@@ -27,9 +27,9 @@
 #' y <- col_relabel(x, Sepal.Length = "Sepal Length of iris flower")
 #' col_labels(y)
 #'
-#' @source This function was taken 1-1 from
+#' @source These functions were taken from
 #' \href{https://cran.r-project.org/package=formatters}{formatters} package, to reduce the complexity of
-#' the dependency tree.
+#' the dependency tree and rewritten.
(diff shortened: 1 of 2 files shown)


R/formatters_var_labels.R
```diff
@@ -38,34 +38,28 @@ col_labels <- function(x, fill = FALSE) {
   checkmate::assert_data_frame(x)
```
gogonzo, line 38: `col_labels` will not work with `DataFrame`. We need to relax this assertion to everything which can have `attr(<column>, "label")`. It was first fixed by @kartikeyakirar and got reverted because of a different reason.
My point is that functions in this file should be more universal and be also applicable to `DataFrame`.

chlebowa: The assertions were causing CI failures and so they were reverted to what can be found in `formatters`. Note that this is now also on `main`.

All right, we need `data.frame` and `DataFrame`. Anything else? Matrices?

gogonzo (approved):
👍 I hope there is no other `data.frame` alike widely used class. We can always add it in the future if needed.

PR #248, opened 2024-01-15
refining tags in documentation

kartikeyakirar:
following discussion on https://github.com/insightsengineering/teal.slice/pull/506#discussion_r1450620331 

I refined the rd documentation by removing the `@description` and` @details` tags,  And using an explicit `@description `tag to prevent the second and subsequent paragraphs from being turned into the `@details`. I followed the guide from https://r-pkgs.org/man.html#title-description-details.

### DIFF
--- a/R/cdisc_data.R
+++ b/R/cdisc_data.R
@@ -1,6 +1,8 @@
 #' Data input for `teal` app
 #'
-#' `r lifecycle::badge("stable")`\cr
+#' @description
+#' `r lifecycle::badge("stable")`
+#'
 #' Function is a wrapper around [teal_data()] and guesses `join_keys`
 #' for given datasets whose names match ADAM datasets names.
 #'
--- a/R/cdisc_data.R
+++ b/R/cdisc_data.R
@@ -1,6 +1,6 @@
 #' Data input for `teal` app
 #'
-#' @description `r lifecycle::badge("stable")`
+#' `r lifecycle::badge("stable")`\cr
 #' Function is a wrapper around [teal_data()] and guesses `join_keys`
 #' for given datasets whose names match ADAM datasets names.
 #'
@@ -12,9 +12,7 @@
 #'
 #' @return A `teal_data` object.
 #'
-#' @details This function checks if there were keys added to all data sets
-#'
-#' @export
+#' @details This function checks if there were keys added to all data sets.
 #'
 #' @examples
 #' data <- cdisc_data(
@@ -27,6 +25,8 @@
 #'   ADSL <- example_cdisc_data("ADSL")
 #'   ADTTE <- example_cdisc_data("ADTTE")
 #' })
+#'
+#' @export
 #'
 cdisc_data <- function(...,
                        join_keys = teal.data::default_cdisc_join_keys[names(rlang::list2(...))],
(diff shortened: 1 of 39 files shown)


R/join_key.R
```diff
@@ -12,9 +11,9 @@
 #' where `names(keys)` maps columns in `dataset_1` corresponding to columns of
 #' `dataset_2` given by the elements of `keys`.
 #'
-#' If unnamed, the same column names are used for both datasets.
+#' * If unnamed, the same column names are used for both datasets.
 #'
-#' If any element of the `keys` vector is empty with a non-empty name, then the name is
+#' * If any element of the `keys` vector is empty with a non-empty name, then the name is
```
averissimo, line 16: Nice to use unordered list for this one :+1: 

Maybe remove empty line between them.

Consistency in markdown: Can you use `-` as below and as the rest of the documentation (either that or replace all with `-`)

````suggestion
#' - If unnamed, the same column names are used for both datasets.
#' - If any element of the `keys` vector is empty with a non-empty name, then the name is
````

**Review by averissimo — APPROVED** [review]
:tada::rocket: It looks good.

I don't like the use of `\cr` on MD, but it's unavoidable as it will consider any additional paragraph as part of `@details` (unless we want to keep more instances of `@description` in addition the the remaining one)

I left a minor comment about markdown list consistency

File: R/cdisc_data.R [hunk 2]
```diff
@@ -1,6 +1,6 @@
 #' Data input for `teal` app
 #'
-#' @description `r lifecycle::badge("stable")`
+#' `r lifecycle::badge("stable")`\cr
```
averissimo, line 3: Might be against the purpose of this PR, but consider this solution to avoid `\cr``_(if it makes sense avoiding `\cr`, it might be a personal thing)_

```suggestion
#' @description `r lifecycle::badge("stable")`
#'
```

kartikeyakirar: Okay, we can add `@description` tags back to the `lifecycle badges`. I also noticed that ` @description` is used wherever badges are used in tidyverse packages. Another alternative would be to remove `\cr` so that it appears without a line break.
image

R/join_key.R
kartikeyakirar: I will replace this with a hyphen (-) as it is used in many places.

R/cdisc_data.R
averissimo: Cool finding on tidyverse, we can copy whatever they are doing :grin: 

It seems like they are going with line break

image

kartikeyakirar: yes, I will add back changes to `teal.slice `as well.

kartikeyakirar: [401a3f9](https://github.com/insightsengineering/teal.data/pull/248/commits/401a3f95632ba46fa332f7d6552f18fc1aafa35e)

chlebowa: I remember one case where I thought about putting in an empty line instead of a break sign. I decided to leave the break because we a) use them in argument description anyway, b) some functions (then) didn't have `@details`, so I wanted to make sure the `@description` is not broken up, and c) no empty line in docs - no empty line in source.

kartikeyakirar: removed other instance of `\cr` [9d07b8c](https://github.com/insightsengineering/teal.data/pull/248/commits/9d07b8ce65a89cbe542a3c6537b612710eec5f23)

PR #254, opened 2024-01-18
Fix `extract_calls` for calls with `;`

m7pr:
Fixes code analysis that has `;` in it. Cases like `code  <- "d <- 'a';assign(value = 'test_text', x = d)"` break `extract_calls()`


chlebowa:
Could you add a unit test that fails without this change?

m7pr:
I even had one in here 

https://github.com/insightsengineering/teal.data/pull/253/files#diff-b02b642a2dfd595132b8d0c2f9a12b1fe8482d68fb58878c9f10dbcf181575f7R167-R176

added with this commit https://github.com/insightsengineering/teal.data/pull/254/commits/f2fd37fbf7c2da5146f38d131b7ed38c8de108ef

chlebowa (approved):

PR #259, opened 2024-01-19
Registers `all.equal.join_keys` only for testing

### LINKED ISSUE #256: `devtools::run_examples` fails
```
r$> devtools::run_examples(
      document = FALSE, 
      run_donttest = TRUE, 
      run_dontrun = FALSE
    )
── Running 36 example files ───────────────────────────────────── teal.data ──
ℹ Loading teal.data
Error in `FUN()`:
! In path: "/Users/ruckip/Documents/repo/gh/insightsengineering/teal.data/tests/testthat/helper-all.equal.R"
Caused by error:
! object 'all.equal.join_keys' not found
Run `rlang::last_trace()` to see where the error occurred.
```

averissimo:
# Pull Request

Fixes #256 

#### Alternative solution

Export the method to compare `join_keys` (method ignores internal list order, NULL elements and will check attributes)

#### Changes description

- Registers S3 method only during testing. `devtools::run_examples()` was sourcing a {testthat} helper file

### DIFF
--- a/tests/testthat/helper-all.equal.R
+++ b/tests/testthat/helper-all.equal.R
@@ -1,2 +1,4 @@
-# covr package doesn't detect the .S3method if it is declared within R/ folder
-.S3method("all.equal", "join_keys", all.equal.join_keys)
+# Register method for testthat::expect_identical to compare join_keys
+if (nzchar(Sys.getenv("TESTTHAT"))) {
+  registerS3method("all.equal", "join_keys", all.equal.join_keys)
+}


chlebowa:
I don't see the method used directly in the package. Is it called by `expect_equal`?

Is there an argument against just exporting the method?

tests/testthat/helper-all.equal.R
```diff
@@ -1,2 +1,4 @@
-# covr package doesn't detect the .S3method if it is declared within R/ folder
-.S3method("all.equal", "join_keys", all.equal.join_keys)
+# Register method for testthat::expect_identical to compare join_keys
+if (nzchar(Sys.getenv("TESTTHAT"))) {
```
pawelru, line 2: have you looked into `testthat::is_testing()` (and possibly other functions like this) 
docs: https://testthat.r-lib.org/reference/is_testing.html

This looks very close to what this function does but the function might also account other edge cases

averissimo: (TIL) I wasn't aware of that function, that is a better solution!

079461a5

averissimo:
> Is there an argument against just exporting the method?

Only that it exports a new function. (I'm not against it, it would actually remove the need for the helper file)

Given that it is only used in `testthat::expect_equal` it wasn't slated to be exported.

chlebowa:
If the function is well made and tested, we may as well export it, maybe someone will have a use for it.

Or we can test with `expect_true(all.equal(jk1, jk2))` instead of `expect_equal`. I believe that is what we do with `teal_slices`.

chlebowa:
Can we estimate if and when that change will be applied?

averissimo:
Nop, the PR has been up since November 20 without any feedback on it _(nor on the issue)_. I don't think it would be fast.

Nevertheless, we would still need to migrate {teal.data} to 3^rd edition. Which should be quick and easy as I already estimated here https://github.com/insightsengineering/NEST-roadmap/issues/65#issuecomment-1816622973

gogonzo:
I think `all.equal` should stay unexported. After release we will remove deprecated and the NAMESPACE will be way smaller. Lets wait with adding more utilities to the new classes.

NAMESPACE after removal od deprecated

Skärmavbild 2024-01-19 kl  15 11 14

gogonzo (approved):
👍

PR #263, opened 2024-01-22
261 multiple improvements to `get_code` function and it's documentation

### LINKED ISSUE #261: [Bug]: get_code doesn't return dependencies of the lines with # @linksto
### What happened?

```r
tdata <- teal_data() |>
  eval_code("
    foo <- function() {
      env <- parent.frame()
      env$x <- 0
    }
    foo() # @linksto x
    y <- x
  ")
  
  
get_code(tdata, datanames = "x")
# character(0)
```

In this case get_code should return all the code

m7pr:
~~Close #216~~

EDIT
Close #261


chlebowa:
Wrong link?

R/teal_data-get_code.R
```diff
@@ -6,9 +6,57 @@
...
-#' see [`get_code_dependency()`].
+#' @section Extracting dataset-specific code:
+#' When `datanames` is specified, the code returned will be limited to the code needed to create the
+#' requested data sets. `code` stored in the `teal_data` object is statically analyzed to determine
+#' dependency tree between each line of the code. Analysis is performed automatically basing on the
+#' used symbols and it is working in a typical case when new dataset is created by assignment operator.
+#' For example:
+#' ```r
+#' data <- teal_data() |>
+#'   eval_code("
```
gogonzo, line 17: @chlebowa I was thinking to use `within` in these examples and use `eval_code` only when `# @linksto` is added (with the note that `within` discards comments)

m7pr: +1

R/teal_data-get_code.R
```diff
@@ -6,9 +6,57 @@
 #' Use `datanames` to limit the code to one or more of the data sets enumerated in `@datanames`.
 #' If the code has not passed verification (with [`verify()`]), a warning will be prepended.
 #'
-#' @section Notes for Developers:
-#' To learn more about how a subset of code needed to reproduce a specific data set is extracted from all code,
-#' see [`get_code_dependency()`].
+#' @section Extracting dataset-specific code:
+#' When `datanames` is specified, the code returned will be limited to the code needed to create the
```
m7pr, line 10:
```suggestion
#' When `datanames` is specified, the returned code will be limited to the code needed to create the
```

R/teal_data-get_code.R
```diff
@@ -6,9 +6,57 @@
...
+#' requested data sets. `code` stored in the `teal_data` object is statically analyzed to determine
+#' dependency tree between each line of the code. Analysis is performed automatically basing on the
```
m7pr, line 12:
```suggestion
#' dependency tree between each line of the code. Analysis is performed automatically based on the
```

R/teal_data-get_code.R
```diff
@@ -6,9 +6,57 @@
...
+#' used symbols and it is working in a typical case when new dataset is created by assignment operator.
```
m7pr, line 13:
```suggestion
#' used symbols and it works in a typical case when data sets are created by the assignment operator.
```

R/teal_data-get_code.R
```diff
@@ -6,9 +6,57 @@
...
+#' data <- teal_data() |>
+#'   eval_code("
+#'     foo <- function(x) x + 1
+#'     x <- 0
+#'     y <- foo(x)
+#'   ")
+#' get_code(data, datanames = "y")
+#' ```
+#'
+#' In above case `y` depends on `z` and `foo` so the code returned by `get_code` will contain all three lines of code.
```
m7pr, line 25:
```suggestion
#' In above case `y` depends on `x` and `foo`, hence the code returned by `get_code` will contain all three lines of code.
```

R/teal_data-get_code.R
```diff
@@ -6,9 +6,57 @@
...
 #'
-#' @section Notes for Developers:
-#' To learn more about how a subset of code needed to reproduce a specific data set is extracted from all code,
-#' see [`get_code_dependency()`].
+#' @section Extracting dataset-specific code:
+#' When `datanames` is specified, the code returned will be limited to the code needed to create the
+#' requested data sets. `code` stored in the `teal_data` object is statically analyzed to determine
+#' dependency tree between each line of the code. Analysis is performed automatically basing on the
+#' used symbols and it is working in a typical case when new dataset is created by assignment operator.
+#' For example:
```
m7pr, line 14:
```suggestion
#' Consider:
```

R/teal_data-get_code.R
```diff
@@ -6,9 +6,57 @@
...
+#'     x <- 0
+#'     y <- foo(x)
+#'   ")
+#' get_code(data, datanames = "y")
+#' ```
+#'
+#' In above case `y` depends on `z` and `foo` so the code returned by `get_code` will contain all three lines of code.
+#' `get_code(data, datanames = "x")` will return only the second line of code etc.
+#' \cr
+#' There could be cases when the dependency tree is not obvious. For example:
```
m7pr, line 28:
```suggestion
#' There could be cases where the dependency tree is not obvious. For example:
```

R/teal_data-get_code.R
```diff
@@ -6,9 +6,57 @@
...
+#'       env <- parent.frame()
+#'       env$x <- 0
+#'     }
+#'     foo()
+#'     y <- x
+#'   ")
+#' get_code(data, datanames = "y")
+#' ```
+#'
+#' In above case `y` depends on `x` but `x` is not created by assignment operator. In such cases
```
m7pr, line 42:
```suggestion
#' In above case `y` depends on `x` but `x` is not created by the assignment operator. In such cases
```

R/teal_data-get_code.R
```diff
@@ -6,9 +6,57 @@
...
+#' `get_code(data, y)` will only return the second line of the code. To overcome this limitation
```
m7pr, line 43:
```suggestion
#' `get_code(data, y)` will only return the last line of the code. To overcome this limitation
```

R/teal_data-get_code.R
```diff
@@ -6,9 +6,62 @@
...
+#'     x <- 0
+#'     y <- foo(x)
+#'   )
+#' get_code(data, datanames = "y")
+#' ```
+#'
+#' In above case `y` depends on `x` and `foo` so the code returned by `get_code` will contain all
+#' three lines of code. `get_code(data, datanames = "x")` will return only the second line of code etc.
+#' \cr
+#' When a code uses non-standard evaluation `get_code` won't be able to determine relationships
```
gogonzo, line 30: WDYT @chlebowa @m7pr

chlebowa: I don't quite understand. Can you provide an example of NSE code that is not analyzed properly?

gogonzo: Good point.
```
{ 
  env <- parent.frame()
  env$x <- 0
}
```
 I took NSE as a term explaining above but it doesn't seem to precise for me. NSE definition is also unwritten. It refers to code substitution and custom scoping rules. I'll try to find better words for this

R/teal_data-get_code.R
```diff
@@ -6,9 +6,62 @@
...
+#'
+#' ```r
+#' data <- teal_data() |>
+#'   eval_code("
+#'     foo <- function() {
+#'       env <- parent.frame()
+#'       env$x <- 0
+#'     }
+#'     foo() # @linksto x
+#'     y <- x
```
chlebowa, line 60: How about this?
```suggestion
#'     x <- 1
#'     foo <- function() {
#'       x <<- 10
#'     }
#'     foo() # @linksto x
#'     y <- x
```

R/teal_data-get_code.R
m7pr: We meant code like that
```r
z <- 'a'
assign(z, 5)
foo <- function(x, value) {
    assign(x, value, envir = .GlobalEnv)
 }
 foo('a', 5)
```

R/teal_data-get_code.R
m7pr: way better and simpler

gogonzo: Haha you got me here! Thanks

R/teal_data-get_code.R
chlebowa: `library(magrittr)` is NSE, as opposed to `library("magrittr"). So is `list$element` as opposed to `lit[["element"]]`. `within` uses NSE as well but `x <- within(iris, Sepal.Length * Sepal.Width)` is processed properly.

chlebowa: > We meant code like that

This is standard evaluation.

gogonzo: Non-standard environment scoping, custom environment scoping - any other propositions to replace NSE?

gogonzo: I'll change this line to: 
> `get_code` isn't always able to properly assess dependencies between each calls and symbols in a provided code.

I will not specify then what exactly causes parser to fail, because the list might be too long and still not exhaustive. What it doesn't support or might fail on:

- updating and reusing objects in a different environment (non standard scoping)
- NSE
- custom assign operators (like `%<>%`) etc.

 I wouldn't specify very technical terms either as 99% of the users will have problems to comprehend it

chlebowa: > `get_code` isn't always able to properly assess dependencies between each calls and symbols in a provided code.

I think this is too broad. Here is a function that detects dependencies between symbols but it doesn't always work.

chlebowa: The reason I asked the initial question is I cannot really say where the mechanism is expected to fail myself. The `env$x` example you provide is so extreme I cannot imagine a use case in the the context of preprocessing data for `teal` applications.

m7pr: Well, maybe we can say it analyzes the code and detects standard relations that result from `<-` operator? Object relations specified in different way may not work, hence we provided `#@linksto` tag to manually specify relations created with other way than `<-` operator.

gogonzo: > analyzes the code and detects standard relations that result from `<-` operator? Object relations specified in different way may not work, hence we provided #@linksto tag to manually specify relations created with other way than `<-` operator.

Fine for me as it highlight what is properly parsed for sure.

gogonzo: @chlebowa any counter proposition?

chlebowa: Well, we do handle `assign` and `data`, and (I think) `<<-` too.

Sorry if this feels like nitpicking. I think I'm missing something: I don't know what the function is unable to do.

gogonzo: @chlebowa 

Not able to determine dependencies on functions which have side effects

```r
data <- teal_data() |> eval_code(
  code = "
    x <- 0
    foo <- function() {
      x <<- x + 1
    }
    foo()
    y <- x
  "
)

identical(
  get_code(data, datanames = "x"),
  paste(
    "x <- 0",
    "foo <- function() {",
    "    x <<- x + 1",
    "}",
    "foo()",
    collapse = "\n"
  )
)
# [1] FALSE
```

not able to track "language" objects
```r
data <- teal_data() |> eval_code(
  code = "
    expr <- quote(x + 1)
    x <- 0
    eval(expr)
  "
)

identical(
  get_code(data, datanames = "x")
  paste(
    "expr <- quote(x + 1)",
    "x <- 0",
    "eval(expr)"
  )
)
# [1] FALSE
```

Doesn't understand options object
```
data <- teal_data() |> eval_code(
  code = '
    options(super_option = TRUE)
    if (getOption("super_option")) {
      x <- 0
    }
  '
)

identical(
  get_code(data, datanames = "x"),
  paste(
    "options(super_option = TRUE)",
    "if (options(\"super_option\")) {",
    "    x <- 0",
    "}",
    collapse = "\n"
  )
)
# [1] FALSE
```

understands when environment is an object (assumess it is a lists)
```r
data <- teal_data() |> eval_code(
  code = '
    y <- 0
    .GlobalEnv$hehe <- y
    x <- .GlobalEnv$hehe
  '
)

identical(
  get_code(data, datanames = "x"),
  paste(
    "y <- 0",
    ".GlobalEnv$hehe <- y",
    "x <- .GlobalEnv$hehe"
  )
)
```

Can't properly determine scoping of the functions
```r
data <- teal_data() |> eval_code(
  code = '
    Sepal.Length <- 0
    x <- dplyr::mutate(iris, yy = Sepal.Length)
  '
)

identical(
  get_code(data, datanames = "x"),
  paste(
    "x <- dplyr::mutate(iris, yy = Sepal.Length)"
  )
  )
)
# [1] FALSE
```

Can't conclude what ifluences environment and that does not
```r
data <- teal_data() |> eval_code(
  code = '
    x <- 1
    assign("x", 0, envir = environment())
    y <- x
  '
)

identical(
  get_code(data, datanames = "y"),
  paste(
    'x <- 1',
    'assign("x", 0, envir = environment())',
    'y <- x'
  )
)
# [1] FALSE
```

gogonzo: Can't relate to for loops
```r
data <- teal_data() |> eval_code(
  code = '
    x <- 10
    y <- list()
    for (xx in 1:x) {
      x[[xx]] <- xx
    }
  '
)

identical(
  get_code(data, datanames = "y"),
  paste(
    "x <- 10",
    "y <- list()",
    "for (xx in 1:x) {",
    "    x[[xx]] <- xx",
    "}",
    collapse = "\n"
  )
)
# [1] FALSE
```

gogonzo: All above can be linked by adding `# @linksto` of course

chlebowa: Thank you, I can work with that :+1:

chlebowa: > understands when environment is an object (assumess it is a lists)

This one actually works, you missed a `sep = "\n"`.

> Can't properly determine scoping of the functions

This one is tricky because the first line is included, where it should be omitted. We would need an ignore tag to handle this, I guess.
This is the only example of NSE, by the way.

> Can't conclude what ifluences environment and that does not

This is actually a bug because `data[["y"]]` ends up being `0` in both cases. @m7pr

R/teal_data-get_code.R
```diff
@@ -6,20 +6,82 @@
...
 #'
 #' @param object (`teal_data`)
 #' @param datanames `r lifecycle::badge("experimental")` (`character`) vector of data set names to return the code for.
 #' @param deparse (`logical`) flag specifying whether to return code as `character` (`deparse = TRUE`) or as
 #' `expression` (`deparse = FALSE`).
 #'
 #' @return
-#' Either string or an expression representing code used to create the requested data sets.
+#' Either a character string or an expression. If `datanames` is used to request a specific dataset,
+#' only code that _creates_ that dataset (not code that uses it) is returned. Otherwise, all contents of `@code`.
```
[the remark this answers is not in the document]
m7pr: CC @chlebowa

R/teal_data-get_code.R
```diff
@@ -6,20 +6,82 @@
...
 #' If the code has not passed verification (with [`verify()`]), a warning will be prepended.
 #'
-#' @section Notes for Developers:
-#' To learn more about how a subset of code needed to reproduce a specific data set is extracted from all code,
-#' see [`get_code_dependency()`].
+#' @section Extracting dataset-specific code:
+#' When `datanames` is specified, the code returned will be limited  to the lines needed to _create_
+#' the requested data sets. The code stored in the `@code` slot is analyzed statically to determine
+#' which lines the datasets of interest depend upon. The analysis works well when objects are created
+#' with standard infix assignment operators (see `?assignOps`) but it can fail in some situations.
```
m7pr, line 13: Wow, didnt know about `?assignOps` existence

R/teal_data-get_code.R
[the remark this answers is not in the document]
chlebowa: The user should know that refers to slots after reading `?teal_data-class` and other documentation.

m7pr:
Excellent work on the documentation @chlebowa - appreciate it!

R/teal_data-get_code.R
m7pr: > > not able to track "language" objects

> There are some bugs here, which we have attempted to tackle but failed.

We didn't scope that initially. I created an issue for that, for the next release https://github.com/insightsengineering/teal.data/issues/270

> > Can't conclude what ifluences environment and that does not

> This is actually a bug because data[["y"]] ends up being 0 in both cases. @m7pr

Yeah, something is off as `assign("x", 0, envir = environment())` should be returned. Will start one more issue to handle that https://github.com/insightsengineering/teal.data/issues/271

R/teal_data-get_code.R
```diff
@@ -3,23 +3,85 @@
...
+#' Here are known cases where manual tagging is necessary:
+#' - non-standard assignment operators, _e.g._ `%<>%`
+#' - objects used as conditions in `if` statements: `if (<condition>)`
+#' - objects used to iterate over in `for` loops: `for(i in <sequence>)`
+#' - creating and evaluating language objects, _e.g._ `eval(<call>)`
+#'
 #'
 #' @param object (`teal_data`)
-#' @param datanames `r lifecycle::badge("experimental")` (`character`) vector of data set names to return the code for.
+#' @param datanames `r lifecycle::badge("experimental")` (`character`) vector of dataset names to return the code for.
```
gogonzo, line 76:
```suggestion
#' @param datanames `r lifecycle::badge("experimental")` (`character`) vector of dataset names to return the code for.
For more details see the "Extracting dataset-specific code" section. 
```

gogonzo: sorry, I didn't add #' in the beginning of the line.

m7pr:
Hey @gogonzo any final words on this one?

This PR initially fixed issues catched at #261 but we also were able to do below during this PR
- enhance documentation of `get_code`
- remove downstream code extraction during `get_code`
- fix #267
- fix #271

chlebowa:
> this is ready to merge

Pending unit tests, of course.

m7pr:
on it

m7pr:
Should be fixed with 3ef0707f85b627e74fd4c0624f63fbf1bcdc8ea8

This PR is an aggregator of multiple other PRs. It occurred one indepdent PR provided features that broke features provided independently in another one :)

R/utils-get_code_dependency.R
```diff
@@ -124,13 +134,36 @@ fix_comments <- function(calls) {
   # If the first token is a COMMENT, then it belongs to the previous call.
   if (length(calls) >= 2) {
     for (i in 2:length(calls)) {
-      if (grepl("@linksto", calls[[i]][1, "text"])) {
+      comments_first <- grepl("@linksto", calls[[i]][1, "text"])
+      comments_second <- grepl("@linksto", calls[[i]][2, "text"]) && calls[[i]][1, "text"] == ""
+      if (comments_first) {
```
gogonzo, line 139:
```suggestion
      comment_idx <- grep("@linksto", calls[[i]][, "text"])
      if (comments_idx) {
```

m7pr: `comments_idx` can be a vector of length > 1, but I see the idea of simplification in here!

gogonzo (approved):
👍 time to go

R/teal_data-get_code.R
m7pr: no worries. I fixed that later

PR #262, opened 2024-01-19
WIP

gogonzo:
closes #261 
Include docs about `get_code(datanames=)`

PR #268, opened 2024-01-23
267 fix `@linksto` tag with `eval()` in last line of the evaluated code for `get_code`

### LINKED ISSUE #267: `get_code` looses comments, hence `# @linksto` tag if this is used in the last line with `eval()`
`eval` calls are not recognized as dependencies, even when explicitly tagged.

```
> data <- teal_data() |> eval_code(
+   code = '
+     expr <- quote(x <- x + 1)
+     x <- 0
+     eval(expr) # @linksto x
+   '
+ )
> get_code(data, datanames = "x") %>% cat
expr <- quote(x <- x + 1)
x <- 0
```

m7pr:
Close #267 

Used @gogonzo solution from this PR to fix this https://github.com/insightsengineering/teal.data/pull/262/files#diff-66a61facc4f5ca86215d63af40e2645b364950439051784916316ef78d5073c8R102-R110

### DIFF
--- a/R/utils-get_code_dependency.R
+++ b/R/utils-get_code_dependency.R
@@ -138,7 +138,7 @@ fix_comments <- function(calls) {
       }
     }
   }
-  Filter(function(x) nrow(x) != 0, calls)
+  Filter(nrow, calls)
 }
 
 # code_graph ----
--- a/R/utils-get_code_dependency.R
+++ b/R/utils-get_code_dependency.R
@@ -99,7 +99,15 @@ find_call <- function(call_pd, text) {
 #' @keywords internal
 #' @noRd
 extract_calls <- function(pd) {
-  calls <- lapply(pd[pd$parent == 0, "id"], get_children, pd = pd)
+  calls <- lapply(
+    pd[pd$parent == 0, "id"],
+    function(parent) {
+      rbind(
+        pd[pd$id == parent, c("token", "text", "id", "parent")],
+        get_children(pd = pd, parent = parent)
+      )
+    }
+  )
   calls <- Filter(Negate(is.null), calls)
   fix_comments(calls)
 }
@@ -130,7 +138,7 @@ fix_comments <- function(calls) {
       }
     }
   }
-  calls
+  Filter(function(x) nrow(x) != 0, calls)
 }
 
 # code_graph ----
--- a/tests/testthat/test-get_code.R
+++ b/tests/testthat/test-get_code.R
@@ -266,6 +266,19 @@ testthat::test_that("@linksto tag indicate affected object if object is assigned
 
 # @linksto ---------------------------------------------------------------------------------------------------------
 
+testthat::test_that("get_code does not break if @linksto is with eval in last line", {
+  code <- c(
+    "expr <- quote(x <- x + 1)",
+    "x <- 0",
+    "eval(expr) #@linksto x"
+  )
+  tdata <- eval_code(teal_data(), code)
+  testthat::expect_identical(
+    get_code(tdata, datanames = "x"),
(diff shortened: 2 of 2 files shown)


chlebowa:
```
data <- teal_data() |> eval_code(
  code = '
   expr <- quote(a <- x + 1)  # @linksto x
   x <- 0
   eval(expr) # @linksto x
 '
)
get_code(data, datanames = "x") %>% cat
```
That's better, thank you.

Is it possible to have `eval(expr)` be included automatically, like with `assign` and `data`?

m7pr:
@chlebowa I'm not sure. We do extract `library()` automatically. For `data()` and `assign()` we only extended the detection so that it understands `data()` and `assign()` as ways of the assignment and they extract those lines if they refer to the object for which you are trying to get the code. For `eval({})` if the object is inside `{}` then this is also tackled. But I would not automatically extract `eval` always, cause it sometimes can refer to different object than the one you are asking the code for.

chlebowa:
Oh, I don't mean always.
```
> data <- teal_data() |> eval_code(
+   code = '
+   expr <- quote(x <- x + 1)
+   x <- 0
+   eval(expr) 
+ '
+ )
> get_code(data, datanames = "x") %>% cat
expr <- quote(x <- x + 1)
x <- 0
```

We know to return the line that creates `expr` but we don't know to return the line where it is evaluated.

m7pr:
I think at some point this could be possible but I would like to avoid trying this out before the very close release. We can try and start and issue to think about it. And currently let's merge this one that fixes an issue where `#@linksto x` tag is in the last line which also contains eval

m7pr:
Created an issue for that in here https://github.com/insightsengineering/teal.data/issues/270

tests/testthat/test-get_code.R
```diff
@@ -266,6 +266,19 @@ testthat::test_that("@linksto tag indicate affected object if object is assigned
 
 # @linksto ---------------------------------------------------------------------------------------------------------
 
+testthat::test_that("get_code does not break if @linksto is with eval in last line", {
```
chlebowa, line 269: Does this cover only `eval` or other functions as well?

gogonzo: Other as well. I put this solution first to fix `foo()` example from `get_code` docs

m7pr: However we can rename the test to "get_code does not break if @linksto is in the last line"

m7pr: Once this is merged to #263 I will do a thorough tests cleanup in there

R/utils-get_code_dependency.R
```diff
@@ -130,7 +138,7 @@ fix_comments <- function(calls) {
       }
     }
   }
-  calls
+  Filter(function(x) nrow(x) != 0, calls)
```
gogonzo, line 141:
```suggestion
  Filter(nrow, calls)
```

m7pr: nice

tests/testthat/test-get_code.R
```diff
@@ -267,6 +267,19 @@ testthat::test_that("detects function usage of the assignment operator", {
...
```
gogonzo, line 270: Maybe clarification like this? 
```suggestion
testthat::test_that("get_code does not break if @linksto is put in the last line", {
  # In some cases R parses comment as a separate expression so the comment is not 
  # directly associated with this line of code. This situation occurs when `eval` is in the last
  # line of the code. Other cases are not known but are highly probable.
```

m7pr: thanks @gogonzo

gogonzo (approved):

PR #276, opened 2024-01-24
251 update vignettes@main

### LINKED ISSUE #251: update vignettes
Since `datanames()` is optional, we reckon some explanation about it's purpose needs to be exposed in vignettes.

_EDIT_ by @chlebowa:
Extend reproducibility vignette by including subsetting code by `datanames` and `# @linksto` tag.

chlebowa:
Closes #251


vignettes/teal-data-reproducibility.Rmd
```diff
@@ -10,34 +12,50 @@ vignette: >
...
+Reproducibility is a primary function of the `qenv` class, which `teal_data` inherits from.
+Every data modification in a `teal_data` object is performed in an encapsulated environment, separate from the global environment.
 
-It is important to note that the reproducibility of this object is limited only to the data-code relationship. Other aspects such as the reliability of the data source, reproducibility of the R session (including package versions), and creation and use of objects from other environments (e.g. `.GlobalEnvironment`) cannot be verified properly by the `teal_data`.
+It is important to note that the reproducibility of this object is limited only to the data-code relationship.
+Other aspects such as the reliability of the data source, reproducibility of the R session (including package versions), and creation and use of objects from other environments (e.g. `.GlobalEnv`) cannot be verified properly by `teal_data`.
+It is advisable to always begin analysis in a new session and run all code that pertains to the analysis within the `teal_data` object.
```
m7pr, line 20: 👍

vignettes/teal-data-reproducibility.Rmd
```diff
@@ -10,34 +12,50 @@ vignette: >
...
 
-## Verification status
+## Verification
+
+### Verification status
+
+Every `teal_data` object has a _verification status_, which is a statement of whether the contents of the `env` can be reproduced by `code`.
```
m7pr, line 26: Not mandatory. field or slot? But this is minor
```suggestion
Every `teal_data` object has a _verification status_ field, which is a statement of whether the contents of the `env` can be reproduced by `code`.

Also, should we use `@env` and `@code` instead of `env` and `code` ? In this case `env` is ambiguous, but if you put `@env` this at least can bring some light on the fact it's a part of `teal_object`
```

vignettes/teal-data-reproducibility.Rmd
```diff
@@ -10,34 +12,50 @@ vignette: >
...
+From this perspective, `teal_data` objects that are instantiated empty are _verified_ but ones instantiated with data and code are _unverified_ because the code need not be reproducible.
```
m7pr, line 27: Very minor
```suggestion
From this perspective, `teal_data` objects that are instantiated empty are _verified_ but ones instantiated with data and code parameters are _unverified_ because the code need not be reproducible.
```

m7pr: `... because the code was not executed and might not be reproducible`?
`... because the code was not executed and might not be able to reproduce objects specified in data`?

vignettes/teal-data-reproducibility.Rmd
chlebowa: I wanted to avoid both, actually, because we don't talk about slots in the other vignette. Perhaps that is a mistake, I do recall slots being mentioned in `teal` vignettes.

vignettes/teal-data-reproducibility.Rmd
```diff
@@ -10,34 +12,50 @@ vignette: >
...
+Obviously, `teal_data` objects instantiated with data only are _unverified_ as well.
 
-`teal_data` objects that are instantiated empty are created as _verified_. Objects can be modified only through `eval_code()` and `within()` functions to ensure that the code is saved in the object and always evaluated in the correct environment. Evaluating code in a `teal_data` object _does not_ change its verification status.
+When evaluating code in a `teal_data` object, the code that is stored is the same as the code that is executed, so it is reproducible by definition.
+Therefore, evaluating code in a `teal_data` object _does not_ change its verification status.
+
+The verification status is always printed when inspecting a `teal_data` object.
+Also, when retrieving code, unverified objects add a warning to the code stating that it has not passed verification.
```
m7pr, line 34: ? unsure if we should put the before `code`
```suggestion
Also, when retrieving the code, unverified objects add a warning to the code stating that it has not passed verification.
```

vignettes/teal-data-reproducibility.Rmd
```diff
@@ -10,34 +12,50 @@ vignette: >
...
 
-## Verification
+### Verification process
 
-`teal_data` objects instantiated with data and code run the risk of the code not being reproducible.
-Such objects will be created in an _unverified_ state. 
-To confirm that the code exactly reproduces the variables stored in the object, one must run the `verify()` function.
-If the verification succeeds, the object's state will be changed to _verified_, otherwise an error will be raised.
-When retrieving code, unverified objects will always add a warning to the code stating that it has not passed verification.
+In order to confirm that the code stored in `teal_data` exactly reproduces the contents of the environment, one must run the `verify()` function.
```
m7pr, line 53: What environment?
```suggestion
In order to confirm that the code stored in `teal_data` exactly reproduces the contents of the `env` environment, one must run the `verify()` function.
```

vignettes/teal-data-reproducibility.Rmd
chlebowa: I meant every operation of code retrieval, I think there should be no article in this case. Perhaps a native English speaker can weigh in?

vignettes/teal-data-reproducibility.Rmd
```diff
@@ -10,34 +12,50 @@ vignette: >
...
+This causes the code to be evaluated and the results to be compared to the contents of the environment.
+If the code- executes without errors and the results are the same as the contents already present in the environment, the verification is successful and the object's state will be changed to _verified_.
```
m7pr, line 55:
```suggestion
If the code executes without errors and the results are the same as the contents already present in the environment, the verification is successful and the object's state will be changed to _verified_.
```

vignettes/teal-data-reproducibility.Rmd
chlebowa: Assuming the reader has read the other vignette, they will know. Do you think that is not a valid assumption? Serious question. How many times should we repeat basic intormation?

vignettes/teal-data.Rmd
```diff
@@ -10,18 +10,21 @@ vignette: >
...
-inside `teal` applications. `teal_data` class inherits from [`qenv`](https://insightsengineering.github.io/teal.code/latest-tag/articles/qenv.html) and is meant to be used for reproducibility purposes. 
+The `teal.data` package specifies the data format used in `teal` applications. 
+The `teal_data` class inherits from [`qenv`](https://insightsengineering.github.io/teal.code/latest-tag/articles/qenv.html) and is meant to be used for reproducibility purposes. 
 
 ## Quick Start
 
-To create a `teal_data` class object, use the `teal_data` function. `teal_data` has a number of methods which allow 
-to set and get relevant information from private class slots.
+To create an object of class `teal_data`, use the `teal_data` function.
+`teal_data` has a number of methods to manage relevant information in private class slots.
```
m7pr, line 19: unsure, maybe rephrase to something like below
```suggestion
`teal_data` has a number of methods to manage relevant information in private slots of the class.
```
or ignore :)

vignettes/teal-data-reproducibility.Rmd
chlebowa: :thinking: There is an implicit assumption that we either evaluate code or we create an object with code. Do you think we should explicitly state that in the latter case the code is not run?

vignettes/teal-data-reproducibility.Rmd
m7pr: Ah sorry. This is explained in Getting Started / Introduction vignette. So no need to put `@` before slots

vignettes/teal-data-reproducibility.Rmd
chlebowa: In order to confirm, one _must_. Otherwise confirmation is impossible.

vignettes/teal-data-reproducibility.Rmd
m7pr: Maybe, but just maybe! someone can assume that code during `teal_data(code = )` creation is executed :)
Only for those I would propose to state this was not executed. But if this is stated in Getting Started / Introduciton then we can avoid it here and leave it as it is.

vignettes/teal-data-reproducibility.Rmd
m7pr: I think this is minor so just wanted to highlight. I believe you have broader understanding on when to put `the`. Just though you  could have missed this one. Closing!

vignettes/teal-data-reproducibility.Rmd
m7pr: I am sorry @chlebowa - I started reading this vignette and then realized we have the second one with basic information. When starting only with this one this was not clear, but no one will start just with this one, but will go through the other one first. So ignore please :) resolving

m7pr (approved):
Hey, those 2 vignettes cover a lot of basic and advanced usage. They are both complement each other in a way where each focus on specific details, but they both created a whole full picture of teal_data object and it's typical ways of using it. Great job explaining that. If only all the software in the world had such thoroughly written documentation

vignettes/teal-data-reproducibility.Rmd
chlebowa: I'm on the fence on this one. Perhaps we should add this information to the `teal_data` help file?

m7pr: Yes, let's add a one liner and you can then omit this suggestion

gogonzo (approved):
Wunderbar 💪

vignettes/teal-data-reproducibility.Rmd
```diff
@@ -67,3 +85,31 @@ data_wrong <- teal_data(
...
+  m <- mtcars
+  head(i)
+})
+cat(get_code(data)) # retrieve all code
+cat(get_code(data, datanames = "i")) # retrieve code for `i`
+```
+
+Note that in when retrieving code for a specific dataset, the result is only the code used to _create_ that dataset, not code that _uses_ is.
+
+## Tracking object dependencies
```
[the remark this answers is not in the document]
chlebowa: Thank you :slightly_smiling_face:

vignettes/teal-data.Rmd
```diff
@@ -10,18 +10,21 @@ vignette: >
...
 
 ```{r, results = 'hide'}
```
vedhav, line 21:
```suggestion
```{r, results = 'hide', message=FALSE}
```

Removes this
Screenshot 2024-01-24 at 9 06 04 PM

vignettes/teal-data-reproducibility.Rmd
```diff
@@ -10,34 +12,50 @@ vignette: >
 
 # Reproducibility of `teal_data` objects
 
-Reproducibility is a primary function of the `qenv` class, which `teal_data` inherits from. Every data modification in `teal_data` object is performed in an encapsulated environment, separate from the global environment.
+Reproducibility is a primary function of the `qenv` class, which `teal_data` inherits from.
+Every data modification in a `teal_data` object is performed in an encapsulated environment, separate from the global environment.
 
-It is important to note that the reproducibility of this object is limited only to the data-code relationship. Other aspects such as the reliability of the data source, reproducibility of the R session (including package versions), and creation and use of objects from other environments (e.g. `.GlobalEnvironment`) cannot be verified properly by the `teal_data`.
+It is important to note that the reproducibility of this object is limited only to the data-code relationship.
```
averissimo, line 18: It would be nice to have an overview of the document's section in the introduction to better guide the vignette

```suggestion
The emphasis on reproducibility enables:

- Verification that the contents of the `teal_data` can be reproduced from code
- Retrieval of the code used to create the `teal_data` object
- Tracking the code necessary to build specific objects

It is important to note that the reproducibility of this object is limited only to the data-code relationship.
```

averissimo (approved):
Looks good!

1 minor comment that shouldn't block the merge.

ps. just like @m7pr said, it would be nice if all documentation was this good

vignettes/teal-data-reproducibility.Rmd
chlebowa: I think this is well covered by the table of contents.

PR #281, opened 2024-01-25
Run verification check when check is `TRUE`

vedhav:
Part of https://github.com/insightsengineering/teal/issues/1069

Please make sure that you also use the `teal` branch `1069-fix-code-warnings-when-data-is-verified@main` for testing the example from the issue.

### DIFF
--- a/R/teal_data.R
+++ b/R/teal_data.R
@@ -52,10 +52,14 @@ teal_data <- function(...,
     if (length(data_objects) > 0 && !checkmate::test_names(names(data_objects), type = "named")) {
       stop("Dot (`...`) arguments on `teal_data()` must be named.")
     }
-    new_teal_data(
+    data <- new_teal_data(
       data = data_objects,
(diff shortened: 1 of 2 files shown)


gogonzo:
@vedhav `check` shouldn't be used anymore - vignette about reproducibility shows only one way to verify object https://insightsengineering.github.io/teal.data/main/articles/teal-data-reproducibility.html

`check` should be deprecated with the link to the migration page https://github.com/insightsengineering/teal/discussions/988

vedhav:
Got it. So, do you recommend creating a `teal_data` wrapper function that we just internally use in teal which has a way to check?

vedhav:
In this case, Shall we add a `r lifecycle::badge("deprecated")` and hint that they have to use `verify` method now?

gogonzo:
closing this PR in favour of
https://github.com/insightsengineering/teal.data/pull/282

PR #282, opened 2024-01-25
deprecate check

gogonzo:
Deprecate `check` argument

### DIFF
--- a/R/teal_data.R
+++ b/R/teal_data.R
@@ -15,7 +15,7 @@
 #' @param code (`character`, `language`) optional code to reproduce the datasets provided in `...`.
 #'  Note this code is not executed and the `teal_data` may not be reproducible
 #'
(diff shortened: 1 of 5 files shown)


R/teal_data.R
```diff
@@ -29,11 +28,21 @@
 teal_data <- function(...,
                       join_keys = teal.data::join_keys(),
                       code = character(0),
-                      check = FALSE) {
+                      check) {
   data_objects <- rlang::list2(...)
   if (inherits(join_keys, "join_key_set")) {
     join_keys <- teal.data::join_keys(join_keys)
   }
+  if (!missing(check)) {
```
averissimo, line 36: How about using the recommended way of deprecating an argument (changing default to `check = deprecated()`)?

image

image

gogonzo:
I've sent a package to win-builder. I'll resubmit to CRAN in about 30min.

PR #299, opened 2024-02-22
minor improvements to label functions

chlebowa:
Adds unit tests for `col_labels<-`.
Slightly changes bodies of `col_labels<-` and `col_relabel` so that they are more similar.
Improved argument checks.

### DIFF
--- a/tests/testthat/test-formatters_var_labels.R
+++ b/tests/testthat/test-formatters_var_labels.R
@@ -49,7 +49,7 @@ testthat::test_that("col_labels<- value names must be same as variable names", {
   iris_df <- utils::head(iris, 2)
   testthat::expect_error(
     col_labels(iris_df) <- stats::setNames(as.character(1:5), toupper(names(iris_df))),
-    "undefined columns selected"
+    "Assertion on 'column names' failed: Must be a permutation of set"
   )
 })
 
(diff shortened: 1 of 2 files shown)


tests/testthat/test-formatters_var_labels.R
```diff
@@ -20,8 +21,72 @@ testthat::test_that("col_labels returns a vector of column names when fill = TRU
...
+testthat::test_that("col_labels<- value accepts named vector", {
+  iris_df <- utils::head(iris, 2)
+  testthat::expect_no_error(col_labels(iris_df) <- stats::setNames(as.character(1:5), names(iris)))
+})
+
+testthat::test_that("col_labels<- value names must be same as variable names", {
+  iris_df <- utils::head(iris, 2)
+  testthat::expect_error(
+    col_labels(iris_df) <- stats::setNames(as.character(1:5), toupper(names(iris_df))),
+    "undefined columns selected"
```
gogonzo, line 52: please make an assertion in the col_labels/col_relabel to avoid this unclear message

gogonzo (commented):
One line in `col_labels` not covered.
 
 Skärmavbild 2024-02-26 kl  09 28 57

chlebowa:
The `stop("labels for variables ... are not character strings")` is old logic that I hadn't changed previously. But this is silly, this should be checked by the setter, shouldn't it?

gogonzo:
> The `stop("labels for variables ... are not character strings")` is old logic that I hadn't changed previously. But this is silly, this should be checked by the setter, shouldn't it?

Hehe yes, I think it doesn't make a bigger sense neither. Why would we check on "get". Probably only because someone/something changed label directly by `attr(<col>, "label")`. I think if somebody/something does it we shouldn't really validate the "class" on get. I allow to remove this line if you agree. If not or if you're unsure please cover with a test

chlebowa:
Rearranged a bit.

gogonzo (approved):
👍

PR #302, opened 2024-03-05
52 standardise optional notation

kartikeyakirar:
part of https://github.com/insightsengineering/nestdevs-tasks/issues/52

### DIFF
--- a/R/cdisc_data.R
+++ b/R/cdisc_data.R
@@ -7,11 +7,10 @@
 #' for given datasets whose names match ADAM datasets names.
 #'
 #' @inheritParams teal_data
-#' @param join_keys (`join_keys`) optional, object or a single (`join_key_set`) object
-#'
-#'   object with datasets column names used for joining.
-#'   If empty then it would be automatically derived basing on intersection of datasets primary keys.
-#'   For ADAM datasets it would be automatically derived.
+#' @param join_keys (`join_keys` or single `join_key_set`)
+#' optional object with datasets column names used for joining.
+#' If empty then it would be automatically derived basing on intersection of datasets primary keys.
+#' For ADAM datasets it would be automatically derived.
 #'
 #' @return A `teal_data` object.
 #'
--- a/R/cdisc_data.R
+++ b/R/cdisc_data.R
@@ -7,9 +7,9 @@
 #' for given datasets whose names match ADAM datasets names.
 #'
 #' @inheritParams teal_data
-#' @param join_keys (`join_keys`) object or a single (`join_key_set`) object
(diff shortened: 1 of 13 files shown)


R/cdisc_data.R
```diff
@@ -7,9 +7,9 @@
 #' for given datasets whose names match ADAM datasets names.
 #'
 #' @inheritParams teal_data
-#' @param join_keys (`join_keys`) object or a single (`join_key_set`) object
+#' @param join_keys (`join_keys`) optional, object or a single (`join_key_set`) object
 #'
-#'   (optional) object with datasets column names used for joining.
+#'   object with datasets column names used for joining.
```
m7pr, line 12: Hey, this does not have sense. You just moved `optional` after the first appearance of any class name?
The meaning of this description was that this is either `join_keys` or either `single join_key_set` and in both cases this is optional.
Right now it looks like `join_keys` is optional, but `single join_key_set` is no longer optional.

I would recommend this to be like below:

```r
@param join_keys (`join_keys` or single `join_key_set`) optional object with datasets column names used for joining
```

R/teal_data.R
```diff
@@ -7,12 +7,12 @@
 #'
 #' @param ... any number of objects (presumably data objects) provided as `name = value` pairs.
 #'
-#' @param join_keys (`join_keys`) object or a single (`join_key_set`) object.
+#' @param join_keys (`join_keys`) optional, object or a single (`join_key_set`) object.
 #'
-#'  (optional) object with dataset column relationships used for joining.
+#'  object with dataset column relationships used for joining.
```
m7pr, line 12: The same as here https://github.com/insightsengineering/teal.data/pull/302/files#r1524668589

R/teal_data.R
```diff
@@ -7,12 +7,12 @@
...
 #'  If empty then no joins between pairs of objects.
 #'
-#' @param code (`character`, `language`) optional code to reproduce the datasets provided in `...`.
+#' @param code (`character`, `language`) optional, code to reproduce the datasets provided in `...`.
```
m7pr, line 15: Since `optional` is an adjective and `code` is a noun it's ok not to have comma between them. Only put comma after adjectives if they come before verbs.
```suggestion
#' @param code (`character`, `language`) optional code to reproduce the datasets provided in `...`.
```

m7pr (changes requested):
Hey, 3 changes and all 3 are improper. Left comments. Please take a look and improve the PR. Then re-request the review from me again.

R/cdisc_data.R
kartikeyakirar: Oh , misread it. thanks. fixed it as per suggestion.

m7pr (approved):

PR #306, opened 2024-04-03
fix order

gogonzo:
Fix order of returned lines

### DIFF
--- a/R/utils-get_code_dependency.R
+++ b/R/utils-get_code_dependency.R
@@ -56,7 +56,7 @@ get_code_dependency <- function(code, names, check_names = TRUE) {
 
   lib_ind <- detect_libraries(calls_pd)
 
-  as.character(code[unique(c(lib_ind, ind))])
+  as.character(code[sort(unique(c(lib_ind, ind)))])
 }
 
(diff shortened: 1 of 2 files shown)


R/utils-get_code_dependency.R
```diff
@@ -56,7 +56,7 @@ get_code_dependency <- function(code, names, check_names = TRUE) {
```
gogonzo, line 56: line 55 loops by dataname - which means that line-idx are grouped by dataname. `sort` few lines later fixes this

kartikeyakirar (approved):
LGTM 🚀 

Tested on simple example.
```

tdata <- teal_data(x1 = iris, x2 = mtcars,
                   code = "x1 <- iris
                   x2 <- mtcars
                   x1 <- x1[1:10,]")

get_code(tdata, datanames = c("x1", "x2"))
```

before: 
`"x1 <- iris"       "x1 <- x1[1:10, ]" "x2 <- mtcars"    `

after:
`"x1 <- iris"       "x2 <- mtcars"     "x1 <- x1[1:10, ]"`

PR #323, opened 2024-08-09
Change the way IDs are created in `teal_data()` constructor

m7pr:
Follow-up after https://github.com/insightsengineering/teal/pull/1302#discussion_r1711376544 this discussion.

Results of tests

```r
══ Results ═══════════════════════════════════════════════════════
Duration: 4.3 s

── Skipped tests (2) ─────────────────────────────────────────────
• This is not urgent and can be ommitted with @linksto tag. (1):
  test-get_code.R:478:3
• We will not resolve this, as this requires code evaluation.
  (1): test-get_code.R:270:3

[ FAIL 0 | WARN 0 | SKIP 2 | PASS 274 ]
```


gogonzo (changes requested):
`id` needs to be random. We made an exception in `teal` to simplify code there only. This fails a `join` method:

```r
q1 <- teal_data(a = 1, code = "a <- 1")
q2 <- teal_data(b = 2, code = "a <- 2")
joined <- join(q1, q2)
get_code(joined) |> cat()
# warning('Code was not verified for reproducibility.')
# a <- 1
```

m7pr:
Yes, Ok! That's why I though that join could break it, but could find the code in join that does that to prove myself wrong : ) Thanks for taking a look

gogonzo:
@m7pr I think we you close it

PR #330, opened 2024-08-30
return ls(@env) if @datanames not specified

gogonzo:
`teal.data::datanames()` should return `ls(@env)` if is not specified. IMO `datanames<-` function shouldn't be supported and determining module's datanames should be done through `module$datanames` only. `teal.data::datanames` has been created only for teal-app to limit datanames. For `teal.data` package it doesn't have any usage at all and should be deprecated in the future. I think that we should use `ls(get_env(<teal_data>))` instead in teal.

### DIFF
--- a/R/join_keys.R
+++ b/R/join_keys.R
@@ -142,7 +142,6 @@ join_keys.teal_data <- function(...) {
 #' join_keys(td)
 `join_keys<-.teal_data` <- function(x, value) {
   join_keys(x@join_keys) <- value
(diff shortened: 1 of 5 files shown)


m7pr:
I was wondering the other day what is the actual point of `teal_data@datanames` if we focus on managing module datanames with a parameter or extra `set_datanames` function. I am leaning toward the removal of this field and function for sure. `ls(get_env(<teal_data>))` could have a useful wrapper `ls_env(teal_data)` in `teal` or in `teal_data`

gogonzo:
Not needed

PR #339, opened 2024-10-13
Fix roxygen2 issue with all/all.equal ambiguity

MichaelChirico:
You are affected by https://github.com/r-lib/roxygen2/issues/1587. Came across your package while looking for a solution, so sharing what I found.

### DIFF
--- a/R/testhat-helpers.R
+++ b/R/testhat-helpers.R
@@ -19,7 +19,8 @@
 #'
 #' @seealso [base::all.equal()]
 #' @keywords internal
-#'
+#' @method all.equal join_keys
+#' @exportS3Method all.equal join_keys
 all.equal.join_keys <- function(target, current, ...) {
   .as_map <- function(.x) {
     old_attributes <- attributes(.x)


MichaelChirico:
I have read the CLA Document and I hereby sign the CLA

MichaelChirico:
recheck

gogonzo:
I don't think either that this affects us.

MichaelChirico:
to clarify I'm referring specifically to this:

https://github.com/insightsengineering/teal.data/blob/b5ccb1d500c648ca7f8ba77058246a280ef2fa2b/man/all.equal.join_keys.Rd#L7

llrs-roche:
From [previous experience](https://github.com/llrs/RGCCA/pull/28#issuecomment-487062507) registering the right S3 method helps with the dispatch, even if it is not exported (no `exportMethods(all)` or `exportMethods(all.equal)`) it must be registered to work well. I'm not familiar enough with `join_keys` and `join_key_set` but it is important on this context. 

I also consider this very minimal (see the last commit) and without issues on a local check.

m7pr:
https://github.com/r-lib/waldo/pull/186 got merged

PR #343, opened 2024-10-16
210 depracate `datanames` in `teal.data::get_code`

m7pr:
Part of 
- https://github.com/insightsengineering/teal.code/issues/210

Companion to
- https://github.com/insightsengineering/teal/pull/1388
- https://github.com/insightsengineering/teal.code/pull/214

# Description

Deprecates the usage of `datanames` parameter in `teal.data::get_code(datanames)` in favour of `teal.code::get_code(names)`.

VBUMP package version locally to something greater than `0.6.1` during testing.

``` r
library(teal.data)
#> Loading required package: teal.code
tdata1 <- within(teal_data(), {a <- 1; b <- 2})
get_code(tdata1)
#> [1] "a <- 1\nb <- 2"
get_code(tdata1, datanames = "a")
#> Warning: The `datanames` argument of `get_code()` is deprecated as of teal.data 0.6.1.
#> ℹ Please use the `names` argument of `teal.code::get_code()` instead.
#> ℹ The deprecated feature was likely used in the teal.data package.
#>   Please report the issue at
#>   <https://github.com/insightsengineering/teal.data/issues>.
#> This warning is displayed once every 8 hours.
#> Call `lifecycle::last_lifecycle_warnings()` to see where this warning was
#> generated.
#> [1] "a <- 1"
```

<sup>Created on 2024-10-17 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


m7pr:
I think it should be OK to just delete this method and make a case in `teal.code::get_code` that if someone uses `datanames` 
parameter than this is treated as `names` parameter and a warning is thrown that you should use `names`. If this method in `teal.data` gets removed, user automatically will use `teal.code::get_code` that will understand `names` and `datanames`.

What do you think @gogonzo ?

R/teal_data-get_code.R
```diff
@@ -108,6 +108,7 @@ setMethod("get_code", signature = "teal_data", definition = function(object, dep
   checkmate::assert_flag(deparse)
 
   code <- if (!is.null(datanames)) {
+    warning("get_code(datanames) was deprecated in teal.data 0.6.1, use get_code(names) instead.", call. = FALSE)
```
m7pr, line 111: I was unlucky with using

```r
lifecycle::deprecate_soft(
  when = " 0.6.1", 
  what = "get_code(datanames)", 
  with = "get_code(names)", 
  always = TRUE
)
```

gogonzo:
> I think it should be OK to just delete this method and make a case in `teal.code::get_code` that if someone uses `datanames` parameter than this is treated as `names` parameter and a warning is thrown that you should use `names`. If this method in `teal.data` gets removed, user automatically will use `teal.code::get_code` that will understand `names` and `datanames`.
> 
> What do you think @gogonzo ?

I'm not sure if we can just remove a method. Some people might use `teal.data::get_code()` and thus they might be surprised and lost (because there will be no info about a move to `teal.code`).

R/teal_data-get_code.R
gogonzo: What was the error?

vignettes/teal-data-reproducibility.Rmd
```diff
@@ -98,7 +98,7 @@ data <- within(teal_data(), {
   head(i)
 })
 cat(get_code(data)) # retrieve all code
-cat(get_code(data, datanames = "i")) # retrieve code for `i`
+cat(get_code(data, names = "i")) # retrieve code for `i`
```
m7pr, line 101: should we also prefix with `teal.code`? or use deprecated `teal.data` (with `datanames`) and allow a warning about the deprecation to be displayed?

```suggestion
cat(teal.code::get_code(data, names = "i")) # retrieve code for `i`
```

m7pr: Alrighty, so we better move to `teal.code::get_code(names` in the vignette

R/teal_data-get_code.R
m7pr: I think I had issues with `deprecate_soft`, but with `deprecate_warn` this is working

```r
> devtools::load_all(".")
ℹ Loading teal.data
Loading required package: teal.code
> library(teal.data)
> tdata1 <- within(teal_data(), {a <- 1; b <- 2})
> get_code(tdata1)
[1] "a <- 1\nb <- 2"
> get_code(tdata1, datanames = "a")
[1] "a <- 1"
Warning message:
The `datanames` argument of `get_code()` is deprecated as of teal.data
0.6.1.
ℹ Please use the `names` argument of `teal.code::get_code()` instead.
ℹ The deprecated feature was likely used in the teal.data package.
  Please report the issue at
  <https://github.com/insightsengineering/teal.data/issues>.
This warning is displayed once every 8 hours.
Call `lifecycle::last_lifecycle_warnings()` to see where this warning was
generated. 
```

m7pr: Pushed the changes using lifecycle package https://github.com/insightsengineering/teal.data/pull/343/commits/3942fecb9027fc6689cae15e6e0d02aa113951d8

R/teal_data-get_code.R
```diff
@@ -108,6 +108,12 @@ setMethod("get_code", signature = "teal_data", definition = function(object, dep
   checkmate::assert_flag(deparse)
 
   code <- if (!is.null(datanames)) {
+    lifecycle::deprecate_warn(
+      when = "0.6.1",
+      what = "teal.data::get_code(datanames)",
+      with = "teal.code::get_code(names)",
+      always = TRUE
+    )
```
m7pr, line 116: This is under `if` condition, and not during the opening of the function becuase `datanames` are always passed as they have a default value. So this would trigger the warning always, even though `datanames` would be NULL

R/teal_data-get_code.R
```diff
@@ -108,6 +108,12 @@ setMethod("get_code", signature = "teal_data", definition = function(object, dep
   checkmate::assert_flag(deparse)
 
   code <- if (!is.null(datanames)) {
+    lifecycle::deprecate_warn(
+      when = "0.6.1",
```
m7pr, line 112: I hope we will vbump the package 0.6.1 due to the close release

m7pr:
Updated the opening comment and made this ready for the review!

R/teal_data-get_code.R
```diff
@@ -39,10 +39,10 @@
 #'     foo()
 #'     y <- x
 #'   })
-#' get_code(data, datanames = "y")
+#' teal.code::get_code(data, names = "y")
```
m7pr, line 42: I'm a bit unsure about those changes, to use `teal.code::get_code` directly. But if we don't do that, old `teal.data::get_code` will throw warnings

m7pr: I wonder how `get_code(data, names = "y")` will work for `data` of class `teal_data`.
It will use the `teal.data::get_code` and will treat `names = "y"` as a part of `...`. But `get_code_dependency` will not be used as it's only activated when `datanames` is not `NULL`. So `names` will be ignored.

That's why I put `teal.code::` prefix so it will use the method for `qenv` where `names` is passed to `teal.code::get_code_dependency`.

We either introduce `names` in `teal.data::get_code` or we prefix examples and documentation with `teal.code::`

gogonzo (changes requested):
There is no point of keeping the old code. `get_code` should just use `teal.code::get_code`.


```r
# method for teal_data
function(object, deparse, datanames = NULL, names = datanames, ...) {
  <deprecate>
  teal.code::get_code(object = object,  deparse = deparse, names = names) 
}
```

Current implementation also assumes that `get_code` method for `teal_data` won't be deprecated (that it will stay) because it warns-deprecate only when `datanames` is specified.

m7pr:
@gogonzo totally! This is way easier and cleaner. Will improve! Thanks

R/teal_data-get_code.R
m7pr: Oh, if we reimplement it this way https://github.com/insightsengineering/teal.data/pull/343#pullrequestreview-2375091411

Then there will be a deprecation warning while using `get_code` on `teal_data` when `teal.data` is attached

m7pr: Maybe deprecation warning is fine, if we remove examples and tests? Then nothing will be displayed in R CMD CHECK as a warning

gogonzo: > Oh, if we reimplement it this way [#343 (review)](https://github.com/insightsengineering/teal.data/pull/343#pullrequestreview-2375091411)
> 
> Then there will be a deprecation warning while using `get_code` on `teal_data` when `teal.data` is attached

Good point. Solution is to change a signature and only run this method when `signature = list(object = "teal_data", datanames = "character")`. Only then this method will be run.

gogonzo: Makes sense. Then `if` is needed, we don't want to include `datanames` in the generic (not worth it).

m7pr:
I see some local errors, that are also on CI/GitHub Actions

```r
> library(teal.data)
> tdata1 <- within(teal_data(), {a <- 1; b <- 2})
> get_code(tdata1)
Error in grDevices::pdf(nullfile()) : too many open devices
Called from: grDevices::pdf(nullfile())
```

`grDevices::pdf``is called in the generic in `teal.code`

R/teal_data-get_code.R
```diff
@@ -103,23 +105,23 @@
...
+      lifecycle::deprecate_warn(
+        when = "0.6.1",
+        what = "get_code()",
+        with = "teal.code::get_code()",
+        always = TRUE
+      )
     }
-  } else {
-    parse(text = paste(c("{", code, "}"), collapse = "\n"), keep.source = TRUE)
+    class(object) <- "qenv"
```
m7pr, line 123: I am not sure this is the cleanest way, but we need a way to call `teal.code::get_code` directly for the `qenv` object. Without that this calls the generic, that uses some `pdf` devices and crashes, and does not extract the code at all.

gogonzo: I guess `CallNextMethod()` is the answer here.

pawelru (changes requested):
Have a read here: https://lifecycle.r-lib.org/reference/deprecated.html
- use `lifecycle::deprecated()` in function formals
- use `lifecycle::is_present()` to check if it's used

NEWS.md
```diff
@@ -1,5 +1,10 @@
 # teal.data 0.6.0.9012
 
+### Deprecation
+
+- `get_code(datanames)` `S4` method parameter for `teal_data()` object was soft deprecated. Use `get_code(names)`.
```
pawelru, line 5:
```suggestion
- soft deprecate `datanames` argument of `get_code()`. Use `names` instead.
```

NEWS.md
```diff
@@ -1,5 +1,10 @@
 # teal.data 0.6.0.9012
 
+### Deprecation
```
pawelru, line 3:
```suggestion
### Breaking changes
```

pawelru, line 3: Are we using a section like that? How it is done (possibly in other packages) up until now? I was about to call it a breaking change but it's soft deprecate so it's not necessarily a breaking type of a change

m7pr: I think we put soft-deprecation in breaking changes as well https://github.com/insightsengineering/teal/blob/main/NEWS.md?plain=1#L80

pawelru: ok so let's follow the convention and call it a breaking change then

R/teal_data-get_code.R
m7pr: Thanks! Nice finding - incorporated in here 7d1c6eebffd989cef1bea7129a1cd2955059bdc5

NEWS.md
m7pr: d2cec0ef4d82b62b9a751c7cf93cb76ea0002f50
thanks Pawel for pointing this out

m7pr:
Thanks @pawelru for pointing out the `deprecated` and `is_present`. Very useful set of tools

I incorporated those functions in this commit 36e0a3305dc3c37aa4356fd79d96bffceff63b7c

R/teal_data-get_code.R
```diff
@@ -103,23 +105,22 @@
...
-    } else {
-      paste(code, collapse = "\n")
+    if (lifecycle::is_present(datanames)) {
+      lifecycle::deprecate_warn(
+        when = "0.6.1",
+        what = "teal.data::get_code(datanames = )",
+        with = "teal.code::get_code(names = )",
+        always = TRUE
+      )
+      checkmate::assert_character(datanames, min.len = 1L, null.ok = TRUE)
```
pawelru, line 121: if you move that if before assert then there won't be a need to repeat the logic

R/teal_data-get_code.R
```diff
@@ -74,14 +74,16 @@
 #' @param object (`teal_data`)
 #' @param datanames `r lifecycle::badge("experimental")` (`character`) vector of dataset names to return the code for.
```
pawelru, line 75: please update this

m7pr: thanks!
https://github.com/insightsengineering/teal.data/pull/343/commits/345f398b08f5847ee6392522406459d8836476e1

R/teal_data-get_code.R
m7pr: https://github.com/insightsengineering/teal.data/pull/343/commits/4ce5c54eac9a97bfbde9513d6eab31ae98b582e9

R/teal_data-get_code.R
```diff
@@ -1,12 +1,12 @@
 #' Get code from `teal_data` object
 #'
-#' Retrieve code from `teal_data` object.
+#' `r lifecycle::badge("deprecated")` Use [`teal.code::get_code()`]. Retrieve code from `teal_data` object.
```
pawelru, line 3: hold on here
are we deprecating the whole method?

m7pr: Well, this method had one extra parameter. Once this parameter is deprecated, we actually use `teal.code::get_code()` and we introduce this parameter in there. So the whole `teal.data::get_code` will be deprecated.

m7pr: I think the logic should be adjusted, so that it warns on the whole usage of the function
```r
      lifecycle::deprecate_warn(
        when = "0.6.1",
        what = "teal.data::get_code()",
        with = "teal.code::get_code()",
        always = TRUE
      )
```

m7pr: and I think we can move out `lifecycle::deprecate_warn` outside of the `lifecycle::is_present(datanames)` so it will be as follows

```r
setMethod("get_code",
  signature = "teal_data",
  definition = function(object, deparse = TRUE, names = NULL, datanames = lifecycle::deprecated(), ...) {
    lifecycle::deprecate_warn(
      when = "0.6.1",
      what = "teal.data::get_code()",
      with = "teal.code::get_code()",
      always = TRUE
    )
    if (lifecycle::is_present(datanames)) {
      names <- datanames
    }

    checkmate::assert_character(names, min.len = 1L, null.ok = TRUE)
    checkmate::assert_flag(deparse)

    callNextMethod(object = object, deparse = deparse, names = names, ...)
  }
)
```

pawelru: That actually makes sense to me! However, I might be too far away from the code base to spot any potential issues. Maybe other devs could help on that?

m7pr: What's your take on this @gogonzo ?

gogonzo: This will throw a warning when `get_code` is used without prefix.

```r
pkgload::load_all("teal.code")
pkgload::load_all("teal.data")

data <- teal_data() |> within(a <- 1)
data |> get_code()
# Warning message:
# `get_code()` was deprecated in teal.data 0.6.1.
```

We don't want to throw this warning whenever `get_code()` is run on the `teal_data` object. Probably you want to throw only when `teal.data::get_code()` is used, but then you need to do `sys.call` or something to detect this - I don't recommend, I'd prefer to just deprecate `datanames` and someday this function will just disappear from the package.

m7pr: Ok @gogonzo I though we want to deprecate the whole function, as it doesn't have any specific purpose right now.
So you see the warning every time you use `get_code` so that you know you should move to teal.code.

m7pr: Ok, I'm having second thoughts. 
Let's just deprecate the parameter right now, as it's safer. And then let's remove the function later, someday

m7pr: Just changed the documentation and the implementation, so only the usage of the parameter is deprecated.
https://github.com/insightsengineering/teal.data/pull/343/commits/5aa5a7f06dd4877c9a03fc260c4e5101ac2a4a1d

gogonzo (approved):
👍

R/teal_data-get_code.R
```diff
@@ -103,23 +105,26 @@
...
+      )
+      names <- datanames
+    }
 
-  if (deparse) {
-    if (length(code) == 0) {
-      code
-    } else {
-      paste(code, collapse = "\n")
+    if (!is.null(names) && lifecycle::is_present(datanames)) {
```
pawelru, line 121: have you seen [`rlang::check_exclusive()`](https://rlang.r-lib.org/reference/check_exclusive.html)?

m7pr: that would be nice

gogonzo: What about this comment @m7pr. Is it resolved?

PR #345, opened 2024-10-23
vbump teal.code

m7pr:
Require higher `teal.code` version after we transitioned `teal.data::get_code` into `teal.code::get_code` and we are using `teal.code::get_code` inside `teal.data` package.


gogonzo (approved):
👍

PR #346, opened 2024-10-28
211 `[.teal_data` S3 method

m7pr:
Companion to https://github.com/insightsengineering/teal.code/pull/216

Introduces `[.teal_data` S3 method for `teal_data` that is based on `[.qenv` but also limits `join_keys`.


llrs-roche (commented):
Once I read in detail the code and documentation I have some feedback.
I think the examples will be very helpful, I am not sure if the current changes on documentation have the desired consequences.
 
More importantly matching base's generic argument and take advantage of the S3 methods used here will probably make developing and maintaining the code easier. 

When I checked locally the examples failed (It could be my setup), but running interactively shows it too (so there might be some undeclared expectations/dependencies):

```r
d> data["a"]
Error en get(x, envir = ns, inherits = FALSE): 
  objeto '[.qenv' no encontrado
d> traceback()
4: get(x, envir = ns, inherits = FALSE)
3: utils::getFromNamespace("[.qenv", "teal.code") at teal_data-extract.R#33
2: `[.teal_data`(data, "a")
1: data["a"]
d> packageVersion("teal.code")
[1] '0.5.0.9011'
d> packageVersion("teal.data")
[1] '0.6.0.9015'
```

R/teal_data-extract.R
```diff
@@ -0,0 +1,38 @@
+#'
+#' @section Subsetting:
+#' `x[names]` subsets objects in `teal_data` environment and limit the code to the necessary needed to build limited
+#' objects.
+#'
+#' @param names (`character`) names of objects included in `teal_subset` to subset
+#' @param x (`teal_data`)
```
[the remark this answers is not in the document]
m7pr: @llrs-roche `teal_data` is described but mostly the argument is called `teal_data` or `data`. In this case it is called `x` because we extend `[.` that uses `x` as the first argument. Funny enough, this `x` is not displayed in the documentation. It is only here to satisfy roxygen2 warnings : )

m7pr:
@llrs-roche 

>   objeto '[.qenv' no encontrado

Any chance you can change your local setup to English : P ?

> Sys.setlocale("LC_ALL","English")

https://stackoverflow.com/questions/16347731/how-to-change-the-locale-of-r

m7pr:
@llrs-roche you need to install `teal.code` from `211_subset@main`, the same as `teal.data` from `211_subset@main`. Looks like you did not had `teal.code` installed from `211_subset@main`, hence you could not find a function `[.qenv`.
This is how we prepare PRs for multiple repositories - if functionality is shared between multiple packages, then you need to install all packages from a specific branch to run and check the code.

Thanks for the attempt!

R/teal_data-extract.R
```diff
@@ -0,0 +1,38 @@
...
+#'
+#' @export
+`[.teal_data` <- function(x, names) {
+  checkmate::assert_class(names, "character")
+  names_in_env <- intersect(names, ls(get_env(x)))
+  if (!length(names_in_env)) {
+    return(teal_data())
+  }
+
+  subset_qenv <- utils::getFromNamespace("[.qenv", "teal.code")
```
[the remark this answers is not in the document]
m7pr: Yeah, I was trying to specifically call method for the parent class :) will give it a shot with `NextMethod()`

R/teal_data-extract.R
```diff
@@ -0,0 +1,38 @@
+#'
+#' @section Subsetting:
+#' `x[names]` subsets objects in `teal_data` environment and limit the code to the necessary needed to build limited
+#' objects.
+#'
+#' @param names (`character`) names of objects included in `teal_subset` to subset
```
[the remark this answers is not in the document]
m7pr: yes, but with `i` it works for `numeric` input, and we only support `character` input :)

R/teal_data-extract.R
```diff
@@ -0,0 +1,38 @@
...
+#' data["a"]
+#' data[c("a", "b")]
+#'
+#' join_keys(data) <- join_keys(join_key("a", "b", "x"))
+#' join_keys(data["a"]) # should show empty keys
+#' join_keys(data["b"])
+#' join_keys(data)["a"] # should show empty keys
+#' join_keys(data)["b"]
+#'
+#' @rdname teal_data
```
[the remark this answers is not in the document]
m7pr: Good catch!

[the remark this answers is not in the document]
m7pr: I googled a bit about `@describeIn` in https://roxygen2.r-lib.org/articles/reuse.html and it looks like it's not recommended and rather designed for a set of functions with the same set of arguments. I would rather stick to what we have right now as it lists S3 method on top of the `@usage` section. 

> In general, I no longer recommend @describeIn because I don’t think the heuristics it uses are as good as a thoughtful hand-crafted summary. If you’re currently using @describeIn, you can general replace it with @rdname, as long as you give some thought to the multiple-function @description.

I think the name of the manual page and the order of functions in `@usage` are taken from the order of files names.
So to have `teal_data` (file `teal_data.R`) to be before `[.teal_data` (teal_data-extract.R file) I needed to rename `teal_data.R` to `teal_data-constructor.R` so it comes before `teal_data-extract.R`. 

Check out the current order of functions and the name of the documentation c01612fdcc780e65984e6c10d2f285bd42041edc
image

R/teal_data-extract.R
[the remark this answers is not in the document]
m7pr: @llrs-roche I had issues calling `teal.code::[.qenv` as this has ticks in the names. It looks like you can call it like below

```r
teal.code:::`[.qenv`
```
but with `:::` 3 colons (even though this is exported from `teal.code`). 
That's why I decided to use `utils::getFromNamespace` so that we do omit a warning in R CMD CHECK that claims you rather not use `:::` in your codebase.

-----

Check out how it looks like with `NextMehod`. I needed to define `names` in the environment of the calling function, and it is not passed to `NextMethod`. This looks a bit weird IMHO. fd4fe27d6757d457987a7d5bd5d4e2ab3dfa266e

Maybe I just don't understand how `NextMethod` works but this is less clear IMHO than 

```r
subset_qenv <- utils::getFromNamespace("[.qenv", "teal.code")
```

R/teal_data-extract.R
```diff
@@ -0,0 +1,49 @@
...
+  # NextMethod works by creating a special call frame for the next method. If no new arguments are supplied, the
+  # arguments will be the same in number, order and name as those to the current method but their values will be
+  # promises to evaluate their name in the current method and environment. Any named arguments matched to ... are
+  # handled specially: they either replace existing arguments of the same name or are appended to the argument list.
+  # They are passed on as the promise that was supplied as an argument to the current environment.
+  # (S does this differently!) If they have been evaluated in the current (or a previous environment) they remain
+  # evaluated. (This is a complex area, and subject to change: see the draft ‘R Language Definition’.)
+
+  names <- names_in_env
+  x <- NextMethod("`[`", x)
```
m7pr, line 45: The usage of `NextMethod` in the opposition to `utils::getFromNamespace("[.qenv", "teal.code")` is discussed in here https://github.com/insightsengineering/teal.data/pull/346#discussion_r1819284910

R/teal_data-extract.R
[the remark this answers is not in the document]
llrs-roche: Nice that the name was fixed. Yes, the Collate field of the DESCRIPTION is precisely whish R files go before and affects also the documentation. The `@describeIn` was a easy suggestion to move the documentation.

llrs-roche (dismissed):
Thanks for the changes. I think this is mostly ready for merging, I just added a petition to have a test for a numeric and NA_character_. 

I think, the workaround with `NextMethod()` come from defining it differently than the generic, but seem easy enough.  

I tried loading teal.code@211_subset@main and running [.qenv exampels still got the same problems `object 'parsed_code'  not found`, so it might be something related to my environment.

R/teal_data-extract.R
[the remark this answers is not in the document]
m7pr: @llrs-roche oh yeah, but we don't want to introduce anything for numeric as `qenv` is basically an `environment` that is not subsettable for numeric as it does not have a natural order (like atomic vectors have).

```r
library(teal.code)
q <- qenv()
q <- eval_code(q, "a <- 1; b <- 2")
q@env[1]
# > Error in q@env[1] : object of type 'environment' is not subsettable
q@env[[1]]
# > Error in q@env[[1]] : wrong arguments for subsetting an environment
q@env[["a"]]
```

Numbers are fine for atomic vectors, or simple S3 objects that naturally use indexes, but I don't think it is the case for the environmemnt. We actually don't want it specifically as we only allow extraction by name, not by numeric index, so that you specifically need to provide the name of the object. With numeric input you can get a different element, depending on the current state of the object : )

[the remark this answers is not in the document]
m7pr: > Sorry, to raise a new thing, could you add a test for numeric subsetting and NA_character_? I know we get the Assertion error, but I want to make sure this is tested/documented for users.

No worries, good idea : )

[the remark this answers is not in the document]
m7pr: Here are extended tests 5f17af01828ed28f7d65bf37088972b6f619af1a

R/teal_data-extract.R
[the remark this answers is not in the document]
m7pr: Ok, `x` is in there. In our case I think `x (teal_data)` should be enought. Especially since it's a page about `teal_data` constructor and a way on how `[` S3 works. This is the style we have all over the place. Let me resolve this conversation

m7pr:
> I tried loading teal.code@211_subset@main and running [.qenv exampels still got the same problems object 'parsed_code'  not found, so it might be something related to my environment.

And you did good. I actually left `teal.code` in unstable state yesterday when I left the office https://github.com/insightsengineering/teal.code/pull/216#issuecomment-2444757712 so it was broken. You got the very same error as I got : )

R/teal_data-extract.R
```diff
@@ -0,0 +1,38 @@
...
+#' join_keys(data["b"])
+#' join_keys(data)["a"] # should show empty keys
+#' join_keys(data)["b"]
+#'
+#' @rdname teal_data
+#'
+#' @export
+`[.teal_data` <- function(x, names) {
+  checkmate::assert_class(names, "character")
+  names <- intersect(names, ls(get_env(x)))
```
gogonzo, line 28: Should extract of non-existing index throw a warning?

R/teal_data-extract.R
```diff
@@ -0,0 +1,38 @@
...
+#' join_keys(data["a"]) # should show empty keys
+#' join_keys(data["b"])
+#' join_keys(data)["a"] # should show empty keys
+#' join_keys(data)["b"]
+#'
+#' @rdname teal_data
+#'
+#' @export
+`[.teal_data` <- function(x, names) {
+  checkmate::assert_class(names, "character")
```
gogonzo, line 27:
```suggestion
  checkmate::assert_character(names)
```

tests/testthat/test-extract.R
```diff
@@ -0,0 +1,42 @@
...
+})
+
+testthat::test_that("`[.` handles names as NA_character_", {
+  data <- teal_data(x = 1, a = 2)
+  testthat::expect_warning(
+    testthat::expect_equal(data[NA_character_], teal_data())
+  )
+})
+
+testthat::test_that("`[.` thorws warnings if names is NULL", {
```
gogonzo, line 22:
```suggestion
testthat::test_that("`[.` throws warning if names is NULL", {
```

tests/testthat/test-extract.R
```diff
@@ -0,0 +1,42 @@
+testthat::test_that("`[.` returns empty teal_data for improper names", {
+  data <- teal_data(x = 1, a = 2)
+  testthat::expect_warning(
+    testthat::expect_equal(data["y"], teal_data())
+  )
+})
+
+testthat::test_that("`[.` handles empty names", {
```
gogonzo, line 8: Handling length of the `names` is handled by `qenv` and this test should be in `teal.code`

R/teal_data-constructor.R
```diff
@@ -21,6 +21,9 @@
 #'
 #' @export
 #'
+#' @name teal_data
+#' @rdname teal_data
+#'
```
gogonzo, line 26: is it necessary? `teal_data()` function has documentation under `?teal_data`

gogonzo (commented):
Minor comments

R/teal_data-extract.R
m7pr: Unsure, maybe it doesn't hurt to include such a warning! Let's include this

llrs-roche:
Feel free to merge but I still have problems with executing the code with the latest update from teal.code@211_subset@main:


```r
d> data <- teal_data()
d> data <- eval_code(data, "a <- 1;b<-2")
Error in parse(text = code_split[[i]], keep.source = FALSE) : 
  <text>:1:12: unexpected symbol
1: a <- 1;b<-2a
               ^
d> data["a"]
✅︎ verified teal_data object
<environment: 0x00000171ebb781d0> [L]
Parent: <environment: devtools_shims>
Warning message:
In `[.teal_data`(data, "a") :
  None of `names` elements exist in `teal_data`. Returning empty `teal_data`.
d> data[c("a", "b")]
✅︎ verified teal_data object
<environment: 0x00000171ea8c9710> [L]
Parent: <environment: devtools_shims>
Warning message:
In `[.teal_data`(data, c("a", "b")) :
  None of `names` elements exist in `teal_data`. Returning empty `teal_data`.
d> 
d> session_info()
─ Session info ───────────────────────────────────────────────────────────────────────
 setting  value
 version  R version 4.4.1 (2024-06-14 ucrt)
 os       Windows 11 x64 (build 22631)
 system   x86_64, mingw32
 ui       RStudio
 language en
 collate  Spanish_Spain.utf8
 ctype    Spanish_Spain.utf8
 tz       Europe/Madrid
 date     2024-10-30
 rstudio  2024.09.0+375 Cranberry Hibiscus (desktop)
 pandoc   3.5 @ C:\\Users\\sanchosl\\AppData\\Local\\Pandoc\\pandoc.exe

─ Packages ───────────────────────────────────────────────────────────────────────────
 !  package     * version    date (UTC) lib source
    backports     1.5.0      2024-05-23 [2] CRAN (R 4.4.0)
    brio          1.1.5      2024-04-24 [2] CRAN (R 4.4.1)
    cachem        1.1.0      2024-05-16 [2] CRAN (R 4.4.1)
    callr         3.7.6      2024-03-25 [2] CRAN (R 4.4.1)
    checkmate     2.3.2      2024-07-29 [2] CRAN (R 4.4.1)
    cli           3.6.3      2024-06-21 [2] CRAN (R 4.4.1)
    curl          5.2.3      2024-09-20 [2] CRAN (R 4.4.1)
    desc          1.4.3      2023-12-10 [2] CRAN (R 4.4.1)
    devtools    * 2.4.5      2022-10-11 [2] CRAN (R 4.4.1)
    digest        0.6.37     2024-08-19 [2] CRAN (R 4.4.1)
    ellipsis      0.3.2      2021-04-29 [2] CRAN (R 4.4.1)
    fastmap       1.2.0      2024-05-15 [2] CRAN (R 4.4.1)
    fs            1.6.4      2024-04-25 [2] CRAN (R 4.4.1)
    glue          1.8.0      2024-09-30 [2] CRAN (R 4.4.1)
    htmltools     0.5.8.1    2024-04-04 [2] CRAN (R 4.4.1)
    htmlwidgets   1.6.4      2023-12-06 [2] CRAN (R 4.4.1)
    httpuv        1.6.15     2024-03-26 [2] CRAN (R 4.4.1)
    later         1.3.2      2023-12-06 [2] CRAN (R 4.4.1)
    lifecycle     1.0.4      2023-11-07 [2] CRAN (R 4.4.1)
    magrittr      2.0.3      2022-03-30 [2] CRAN (R 4.4.1)
    memoise       2.0.1      2021-11-26 [2] CRAN (R 4.4.1)
    mime          0.12       2021-09-28 [2] CRAN (R 4.4.0)
    miniUI        0.1.1.1    2018-05-18 [2] CRAN (R 4.4.1)
    pkgbuild      1.4.4      2024-03-17 [2] CRAN (R 4.4.1)
    pkgload       1.4.0      2024-06-28 [2] CRAN (R 4.4.1)
    processx      3.8.4      2024-03-16 [2] CRAN (R 4.4.1)
    profvis       0.4.0      2024-09-20 [2] CRAN (R 4.4.1)
    promises      1.3.0      2024-04-05 [2] CRAN (R 4.4.1)
    ps            1.8.0      2024-09-12 [2] CRAN (R 4.4.1)
    purrr         1.0.2      2023-08-10 [2] CRAN (R 4.4.1)
    R6            2.5.1      2021-08-19 [2] CRAN (R 4.4.1)
    Rcpp          1.0.13     2024-07-17 [2] CRAN (R 4.4.1)
    remotes       2.5.0      2024-03-17 [2] CRAN (R 4.4.1)
    rlang         1.1.4      2024-06-04 [2] CRAN (R 4.4.1)
    rprojroot     2.0.4      2023-11-05 [2] CRAN (R 4.4.1)
    rstudioapi    0.17.1     2024-10-22 [2] CRAN (R 4.4.1)
    sessioninfo   1.2.2      2021-12-06 [2] CRAN (R 4.4.1)
    shiny         1.9.1      2024-08-01 [2] CRAN (R 4.4.1)
    teal.code   * 0.5.0.9012 2024-10-30 [1] Github (insightsengineering/teal.code@a32939c)
 VP teal.data   * 0.6.0.9015 2024-10-21 [?] Github (insightsengineering/teal.data@b5ccb1d) (on disk 0.6.0.9012)
    testthat    * 3.2.1.1    2024-04-14 [2] CRAN (R 4.4.1)
    urlchecker    1.0.1      2021-11-30 [2] CRAN (R 4.4.1)
    usethis     * 3.0.0      2024-07-29 [2] CRAN (R 4.4.1)
    vctrs         0.6.5      2023-12-01 [2] CRAN (R 4.4.1)
    withr         3.0.1      2024-07-31 [2] CRAN (R 4.4.1)
    xtable        1.8-4      2019-04-21 [2] CRAN (R 4.4.1)

 [1] C:/Users/sanchosl/R-dev
 [2] C:/Users/sanchosl/AppData/Local/R/win-library/4.4
 [3] C:/Program Files/R/R-4.4.1/library

 V ── Loaded and on-disk version mismatch.
 P ── Loaded and on-disk path mismatch.
```

R/teal_data-constructor.R
m7pr: Ah! Yes, this is not necessary, I think this is the remaining of some other tests that I did when I included `[.teal_data` in the docs of `teal_data`. Removed in here f74635a762876a39997a678b8e7fb85f1c857f88

R/teal_data-extract.R
pawelru: should we throw a warning at all?

I'm looking at this like a subsetting from env:
```
r$> e <- list2env(list(a = 1, b = 2))

r$> e[['a']]
[1] 1

r$> e[['aa']]
NULL
```
I see no warning in this.

R/teal_data-constructor.R
```diff
@@ -21,6 +21,7 @@
 #'
 #' @export
 #'
+#'
```
m7pr, line 24:
```suggestion
```

R/teal_data-extract.R
gogonzo: @pawelru Marcin is implementing `[` instead of `[[`. I proposed a warning because `[` returns strange results for inexisting indexes which can't be translated one-to-one to `teal_data` object:

```r
e <- list(a = 1, b = 2) # list because environment has no `[` method
e[NULL]
# named list()

e[character(0)]
# named list()
e["c"]

e[]
# $a
# [1] 1
# 
# $b
# [1] 2

e["inexisting"]
# $<NA>
# NULL
```

We can discuss whether to warn or not here, but this should be implemented in `[.qenv`

tests/testthat/test-extract.R
m7pr: Alrighty, moved to teal.code

R/teal_data-extract.R
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  if (length(names_warn)) {
+    warning(
+      sprintf(
+        "Some elements of `names` do not exist in `teal_data`. Skipping those: %s.",
+        paste(names_warn, collapse = ", ")
+      )
+    )
+  }
```
pawelru, line 43: This is almost identical to the method implemented for parent `qenv` class in `teal.code`. The only difference is the class name in warning. Let's keep it DRY. Note that you have access to `x` so you can get its class name if this is really needed and put that in the warning message.

llrs-roche: It might have changes since last time I checked. The code can be simplified to just `x <- NextMethod("[", x)` and the rest of the method.
Accessing the class of the object will be helpful if it is handled on `[.qenv`  it might simplify things if
something like ` warning("None of 'names' elements exist in '", class(x)[1], "'. Returning empty '", class(x)[1], "'.")` is used .

m7pr: thanks, make sense guys

R/teal_data-extract.R
```diff
@@ -0,0 +1,32 @@
...
+#' join_keys(data["b"])
+#' join_keys(data)["a"] # should show empty keys
+#' join_keys(data)["b"]
+#'
+#' @rdname teal_data
+#'
+#' @export
+`[.teal_data` <- function(x, names) {
+  names_no_raw <- names
+  names <- c(names, ".raw_data")
```
gogonzo, line 28: This is reverse dependency. `teal.data` shouldn't know anything about the `.raw_data`.

m7pr: Alrighty, I will extend `[.qenv` so it allows to pass hidden names
`possible_names <- ls(get_env(x))` -> `possible_names <- ls(get_env(x), all.names = TRUE)`

Will remove `.raw_data` it in `teal.data` and will add `.raw_data` to the names in `teal`

m7pr: - Extended in teal.code https://github.com/insightsengineering/teal.code/pull/216/commits/26efa3fdc6bf42b32c422dba20b3a85d3ce092ab
- Reverted in teal.data https://github.com/insightsengineering/teal.data/pull/346/commits/d629a043d0ab882a388d7375c5f2daa93ca27545
- Updated in teal https://github.com/insightsengineering/teal/pull/1396/commits/e2352b2e025daf89a56533e34f7560586168bd28

m7pr:
Since https://github.com/insightsengineering/teal.code/pull/216 is ready, I open this PR for the review. Let's see results of CI/CD checks

R/teal_data-class.R
```diff
@@ -87,12 +91,34 @@ new_teal_data <- function(data,
...
+  if (length(code)) {
+    lapply(split_code(code), function(current_code) {
+      attr(current_code, "id") <- sample.int(.Machine$integer.max, 1)
+      parsed_code <- parse(text = trimws(current_code), keep.source = TRUE)
+      attr(current_code, "dependency") <- extract_dependency(parsed_code)
+      current_code
+    })
+  } else {
+    list(character(0))
+  }
```
gogonzo, line 123: Empty `qenv` has a `list()`. Otherwise `@code` will have a length == 1 
```suggestion
  } else {
    list()
  }
```

R/teal_data-class.R
```diff
@@ -87,12 +91,34 @@ new_teal_data <- function(data,
...
+#' @return list of `character`s of the length equal to the number of calls in `code`.
+#'
+#' @keywords internal
+#' @noRd
+code2list <- function(code) {
+  checkmate::assert_character(code, null.ok = TRUE)
+  if (length(code)) {
+    lapply(split_code(code), function(current_code) {
+      attr(current_code, "id") <- sample.int(.Machine$integer.max, 1)
+      parsed_code <- parse(text = trimws(current_code), keep.source = TRUE)
```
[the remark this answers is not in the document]
gogonzo: Not needed

R/teal_data-class.R
m7pr: I am rewriting this a bit so it will also work for `""` passed as input, will rewrite this anyway. Thanks!

R/teal_data-extract.R
```diff
@@ -0,0 +1,33 @@
...
+#' join_keys(data["a"]) # should show empty keys
+#' join_keys(data["b"])
+#' join_keys(data)["a"] # should show empty keys
+#' join_keys(data)["b"]
+#'
+#' @rdname teal_data
+#'
+#' @export
+`[.teal_data` <- function(x, names) {
+  x <- NextMethod("`[`", x, check_names = FALSE) # takes 'names' from function's environment
```
m7pr, line 27: `check_names = FALSE`

because

```r
td <- teal_data(a = 1, b = 2)
```

has no code, and 

```r
td['a']
```

raises a warning that `a` is missing in the code when extracting code

gogonzo: It should throw a warning. One tries to extract `a` which is not present in the code.

gogonzo: Or maybe a question, WDYT about this? Should it throw or not?

m7pr: Maybe we can have an error on `teal_data` level, that warns about it but with custom warning for teal_data curated for this siu. Without `check_names = FALSE` it gives unreasonable warnning from `get_code_dependency`.

R/teal_data-extract.R
```diff
@@ -0,0 +1,33 @@
...
+  if (inherits(x, "qenv")) {
+    return(teal_data())
+  } # all 'names' not in object
```
gogonzo, line 30: I always inherits from qenv
```suggestion
```

R/teal_data-extract.R
m7pr: This is passed to `get_code` -> `get_code_dependency` and it checks for objects within code
https://github.com/insightsengineering/teal.code/pull/216/files#diff-66a61facc4f5ca86215d63af40e2645b364950439051784916316ef78d5073c8R34

R/teal_data-extract.R
m7pr: Thanks! I needed an exit condition in case empty `qenv` is returned in `teal.code:::[.qenv`
I changed it to this https://github.com/insightsengineering/teal.data/pull/346/commits
Now teal works like a charm

tests/testthat/test-extract.R
```diff
@@ -0,0 +1,39 @@
...
+
+testthat::test_that("`[.` returns limited join_keys", {
+  data <- teal_data(a = 1, b = 2)
+
+  join_keys(data) <- join_keys(join_key("a", "b", "x"))
+  testthat::expect_equal(
+    join_keys(data["a"]),
+    join_keys()
+  )
+})
```
gogonzo, line 39: Some tests to consider.

```suggestion
})

testthat::test_that("`[.` throws when inexisting names", {
  data <- teal_data(x = 1, a = 2)
  testthat::expect_warning(data[c("a", "b")])
})

testthat::test_that("`[.` returns code of inexisting data", {
  data <- teal_data(x = 1, a = 2, code = "c <- 3")
  data["c"] # what now?
})

testthat::test_that("`[.` returns code of inexisting data", {
  data <- teal_data(x = 1, a = 2, code = "a <- 1; c <- 3") # this throws a warning!
  data[c("a", "c")] # what now?
})

testthat::test_that("`[.` preserves @verified field", {
  testthat::expect_false(teal_data(a = 1, b = 2)["a"]@verified)
  testthat::expect_true(within(teal_data(), a <- 1)["a"]@verified)
})

```

gogonzo (approved):
🔥 🧯

PR #347, opened 2024-10-28
Adds `names()` function and deprecates `datanames()`

### LINKED ISSUE #333: [Feature Request]: Don't use @datanames anymore
### Feature description

`@datanames` slot has been introduced only to distinguish temporary/proxy objects from "real" datanames. It was useful in `teal` where app didn't have to create a filter-panel. Now (after [this](https://github.com/insightsengineering/teal/pull/1334)) `teal` ignores `teal.data::datanames` and simply runs `ls(x@env)` to determine "datanames".

We can keep `datanames` function but `names` (implemented in `qenv`) is more relevant in my opinion.

```r
names.qenv <- function(x) ls(x@env)
names(teal_data(a = 1, .b = 2))
#> [1] "a"
```

Plan:
- introduce `names` method for `qenv` and `qenv.error`
- remove `@datanames` slot
- deprecate `datanames<-` in `teal.data` - function has no effect in teal anymore. Please point to use `.` for exclusions or `teal::set_datanames`
- deprecate `datanames()` in `teal.data` - point to the `names`
- replace `teal.data::datanames` in the modules in favour of `names()` 
  - tmc, tmg, teal.gallery, tlg-catalog

Related issues:
* https://github.com/insightsengineering/teal.data/issues/338
* https://github.com/insightsengineering/teal/issues/1366
* https://github.com/insightsengineering/teal/issues/1352
* https://github.com/insightsengineering/teal/issues/1347
* https://github.com/insightsengineering/teal/issues/1378
* https://github.com/insightsengineering/teal/issues/1379
* https://github.com/insightsengineering/teal.code/issues/210
* https://github.com/insightsengineering/teal.code/issues/211
* https://github.com/insightsengineering/teal/issues/1380

averissimo:
# Pull Request

- Fixes #333
- Companion of https://github.com/insightsengineering/teal.code/pull/218

#### Blockers

- `{teal.code}`
    - [ ] Merge https://github.com/insightsengineering/teal.code/pull/218

#### Changes description

- [x] Adds `names()` function
    - [x] Setter is implemented but throws warning (does nothing)
- [x] Deprecates `datanames()` getter and setter
- [x] Adapt tests to new API
- [x] Update NEWS
- [ ] Review documentation
- [ ] Review vignettes


R/deprecated.R
```diff
@@ -343,3 +343,51 @@ get_join_keys <- function(...) {
...
+    "`datanames<-`()",
+    details = "invalid to use `datanames()<-` or `names()<-` on an object of class `teal_data`. See ?names.teal_data"
+  )
+  names(x)
+}
+
+#' @rdname datanames
+#' @export
+#' @keywords internal
+`names<-.teal_data` <- function(x, value) {
```
gogonzo, line 386: I acknowledge introducing `names<-` just to deprecate itself

R/teal_data-names.R
```diff
@@ -0,0 +1,41 @@
...
+#' td <- teal_data(iris = iris)
+#' td <- within(td, mtcars <- mtcars)
+#' names(td)
+#'
+#' td <- within(td, .CO2 <- CO2)
+#' names(td)
+#'
+#' @export
+names.teal_data <- function(x) {
+  names_x <- names(as.environment(x))
```
averissimo, line 24: Using `names` in body which will return all objects, including with dot prefix.

I don't think it makes sense to use `ls(x, all.names = FALSE)` as that changes the way the function works.

We could add support to `ls` via some trickery, but I'm leaning against it, mainly as it will appear as masking `ls` when loading the package.

```r
#' @export
ls <- function(x, ...) {
  UseMethod("ls")
}

#' @export
ls.teal_data <- function(x, ...) {
  names_x <- base::ls(as.environment(x), ...)
  .get_sorted_names(names_x, join_keys(x), as.environment(x))
}

#' @export
ls.default <- function(x, ...) {
  base::ls(x, ...)
}
```

gogonzo: I agree that `ls` would be the best. Does overriding need to come with a warning? I remember some big package was overriding `base::setequal,setdiff,intersect` and I remember we did the same with one package. It wasn't the problem with `base` directly but with other packages which were trying to do the same. Problem with `ls` is that there is `ls.str` so if you create an S3 this might be dispatched and fail because `str` is not a class. The only way is to use S4 I guess.

---

IMO overriding `names.qenv` makes sense even if not exactly the same as `names.environment`. With `qenv` we introduce a new class and methods in some cases can differ from the inherited ones. If we consider using `qenv` object in modules or it teal internals, names for me seems to be the most relevant.

averissimo: Regarding `ls.str` we either need to treat this as protected creating an exception in dispatch, or try to restrict dispatch only to current package namespace

R/teal_data-class.R
```diff
@@ -53,18 +54,11 @@ setClass(
...
 #' @param join_keys (`join_keys`) object
-#' @param datanames (`character`) names of datasets passed to `data`.
-#'   Needed when non-dataset objects are needed in the `env` slot.
 #' @rdname new_teal_data
 #' @keywords internal
-new_teal_data <- function(data,
-                          code = character(0),
-                          join_keys = join_keys(),
-                          datanames = names(data)) {
+new_teal_data <- function(data, code = character(0), join_keys = join_keys()) {
```
gogonzo, line 59: Let's deprecate and not remove argument for now. 
```suggestion
new_teal_data <- function(data, code = character(0), join_keys = join_keys(), datanames = lifecycle::deprecated()) {
```

gogonzo (approved):
I exhausted all comments here. Accept with a minor comment which you can address or not 👍

.pre-commit-config.yaml
```diff
@@ -20,7 +20,7 @@ repos:
           - methods
           - rlang
           - stats
-          - teal.code
+          - insightsengineering/teal.code
```
gogonzo, line 23: I think you need to revert this

```suggestion
          - teal.code
```

averissimo: Ok, I'll start a PR to re-add this and discuss it.

`pre-commit` is a developer tool and it fails to build documentation when there are docs dependencies between packages (for example `@inheritParams`)

R/teal_data-class.R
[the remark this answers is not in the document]
averissimo: Done.

I simplified the constructor and moved logic downstream to `teal.code` as it doesn't make sense to have it on `teal.data`

I believe this is a safe change, as previous version of `teal.code::qenv` class allowed setting `code` and other arguments on `methods::new`.

``` r
library(teal.code)
packageVersion("teal.code")
#> [1] '0.5.0.9012'
q <- methods::new("qenv")
q@env
#> <environment: 0x6527e983ef78>

q <- methods::new("qenv", code = "")
#> Error in validObject(.Object): invalid class "qenv" object: @code and @id slots must have the same length.

q <- methods::new("qenv", code = "", messages = "", warnings = "", id = 1L)
q
#> <environment: 0x6527e983ef78>
#> Parent: <environment: package:stats>
```

<sup>Created on 2024-11-06 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

``` r
pkgload::load_all("teal.code")
#> ℹ Loading teal.code

packageVersion("teal.code")
#> [1] '0.5.0.9012'
q <- methods::new("qenv")
q@env
#> Error: no slot of name "env" for this object of class "qenv"

q <- methods::new("qenv", code = "")
q
#> <environment: 0x591e8c6c5760> [L]
#> Parent: <environment: devtools_shims>

q <- methods::new("qenv", code = "", messages = "", warnings = "", id = 1L)
q
#> <environment: 0x591e8d252f80> [L]
#> Parent: <environment: devtools_shims>
```

<sup>Created on 2024-11-06 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

R/teal_data-class.R
```diff
@@ -14,85 +14,61 @@ setOldClass("join_keys")
...
+setMethod(
+  "initialize",
+  "teal_data",
+  function(.Object, .xData = list(), join_keys = join_keys(), ...) { # nolint: object_name.
+    # Allow .xData to be a list and convert it to an environment
+    if (!missing(.xData) && inherits(.xData, "list")) {
+      .xData <- rlang::env_clone(list2env(.xData), parent = parent.env(.GlobalEnv)) # nolint: object_name.
+      lockEnvironment(.xData, bindings = TRUE)
+    }
+    checkmate::assert_environment(.xData)
```
gogonzo, line 63: I can't make a suggestion because GH doesn't get the differences properly. So I'm just attaching the code of the initialize. I'm just suggesting to utilise `callNextMethod` instead of adding `.Object@` afterwards 

```r
setMethod(
  "initialize",
  "teal_data",
  function(.Object, .xData = list(), join_keys = join_keys(), ...) { # nolint: object_name.
    # Allow .xData to be a list and convert it to an environment
    if (!missing(.xData) && inherits(.xData, "list")) {
      .xData <- rlang::env_clone(list2env(.xData), parent = parent.env(.GlobalEnv)) # nolint: object_name.
      lockEnvironment(.xData, bindings = TRUE)
    }
    args <- list(...)
    checkmate::assert_environment(.xData)
    checkmate::assert_class(join_keys, "join_keys")
    checkmate::assert_list(args, names = "named")
    methods::callNextMethod( # nolint: object_name.
      .Object,
      .xData,
      join_keys = join_keys,
      verified = (length(args$code) == 0L && length(.xData) == 0L),
      ...
    )
  }
)
```

averissimo: It makes sense, this way we delegate the actual slot assignments to the internals of S4 chain of `initialize()`

Done in 04fd26ef

PR #352, opened 2024-11-11
Update `teal.code` version

m7pr:
`[.qenv` was introduced in `teal.code` `0.5.9014` so we need to depend on this or higher to get `[.teal_data`


gogonzo (approved):
👍

PR #354, opened 2024-11-13
id as names

gogonzo:
closes https://github.com/insightsengineering/teal.code/issues/230
alternative to https://github.com/insightsengineering/teal.code/pull/231


gogonzo:
> Wow, that didn't require a huge overhaul/refactor. Probably the docs would need to get changed as well in a couple of places. Maybe let's grep for @id? I think the class definition document should contain information about it

Thanks for highlighting docs. I've figured out that we can just inheritSection about Code

m7pr (approved):

PR #359, opened 2025-01-03
update testthat version

donyunardi:
We need to bump up `testthat` version because we're using `testthat::it` function in unit tests.

This function was introduced in `3.1.8`:
https://github.com/r-lib/testthat/blob/v3.1.8/R/describe.R#L89

This update will fix the `Dependency Test - min isolated`:
https://github.com/insightsengineering/teal.data/actions/runs/12532231366/job/34950746107#step:9:774

### DIFF
--- a/.github/workflows/scheduled.yaml
+++ b/.github/workflows/scheduled.yaml
@@ -36,6 +36,7 @@ jobs:
       GCHAT_WEBHOOK: ${{ secrets.GCHAT_WEBHOOK }}
     with:
       strategy: ${{ matrix.test-strategy }}
+      extra-deps: waldo (>= 0.5.3)
       additional-env-vars: |
         PKG_SYSREQS_DRY_RUN=true
   branch-cleanup:
--- a/.github/workflows/scheduled.yaml
+++ b/.github/workflows/scheduled.yaml
@@ -36,7 +36,6 @@ jobs:
       GCHAT_WEBHOOK: ${{ secrets.GCHAT_WEBHOOK }}
     with:
       strategy: ${{ matrix.test-strategy }}
-      extra-deps: waldo (>= 0.5.3)
       additional-env-vars: |
         PKG_SYSREQS_DRY_RUN=true
   branch-cleanup:
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -40,7 +40,7 @@ Imports:
 Suggests:
     knitr (>= 1.42),
     rmarkdown (>= 2.23),
-    testthat (>= 3.1.8),
+    testthat (>= 3.2.2),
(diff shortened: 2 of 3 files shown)


donyunardi:
Got another failure about `compare_proxy` which seems to be related to `waldo` package:
https://github.com/insightsengineering/teal.data/actions/runs/12605104075/job/35133232553#step:9:235

Updating testthat to 3.2.0

donyunardi:
Still fails:
https://github.com/insightsengineering/teal.data/actions/runs/12605422042/job/35134030219#step:9:853

This may have something to do with the `default_cdisc_join_keys` object.

donyunardi:
Issue still persist. I think it may have something to do with other dependencies based on the min_isolated strategy.

@averissimo @m7pr 
Any chance you can shed a light on this based on your experience working with verdepcheck?
For sure we need to update the testthat version to 3.1.8 but not sure how to address this `compare_proxy` issue.

averissimo:
I'm not able to reproduce the error on my machine (locally with building from source) as verdepcheck never resolves down to 3.1.8 (using 3.2.2 with same strategy)

I'm testing with the CI virtual image and will update this issue when possible, it takes some time on every run :sweat_smile:

llrs-roche:
I was also checking this issue: `compare_proxy` comes from waldo which introduced it very early, but since its origin it has been adding different methods. S4 methods appeared on early, but there have been improvements on its methods as recently as November. 

Latest testthat (December 2024) increased the dependency version of waldo to a waldo version with improvements on S4 methods of `compare` (also November 2024), so we might require a high waldo version (despite what testthat says).

averissimo:
Both solutions work, I'm leaning towards bumping `{testthat}` to `v3.2.2` as that solves the problem cleanly

- `extra-deps` : [CI run](https://github.com/insightsengineering/teal.data/actions/runs/12655213196/job/35264976400)
- Bumping version: [CI run](https://github.com/insightsengineering/teal.data/actions/runs/12655400676/job/35265604684)

image

Note: `release` and `max` fails are expected (unreleased version and CRAN 0-day old package respectively)

donyunardi:
Thanks for looking into this @averissimo @llrs-roche 
I would prefer to update this to 3.2.2.

@averissimo 
Do you think we still need my update to the unit test? Or should we revert this?
https://github.com/insightsengineering/teal.data/blob/5eddd517f598fd702ecc9ce9764a579a65718ffb/tests/testthat/test-cdisc_join_keys.R#L2

averissimo:
That is a great addition here @donyunardi _(today I learned about that parameter)_

averissimo (approved):

PR #362, opened 2025-01-17
fix deprecation message; add renaming to the examples

pawelru:
before:
```
r$> lifecycle::deprecate_soft("0.6.1", "datanames()", details = "names()")
Warning message:
`datanames()` was deprecated in <NA> 0.6.1.
ℹ names()
```

after:
```
r$> lifecycle::deprecate_soft("0.6.1", "datanames()", with = "names()")
Warning message:
`datanames()` was deprecated in <NA> 0.6.1.
ℹ Please use `names()` instead.
```

Also updated docs to give example of reassigning names - previously done with `datanames(x) <- ...` which is deprecated now

### DIFF
--- a/R/deprecated.R
+++ b/R/deprecated.R
@@ -19,7 +19,11 @@
 #' @rdname datanames
 #' @export
 datanames <- function(x) {
-  lifecycle::deprecate_soft("0.6.1", "datanames()", details = "names()")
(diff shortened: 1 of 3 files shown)


llrs-roche (approved):
Looks good. Nice addition of the renaming example, I had to double check the last example as I expected `names(td) <- ` to work.

PR #371, opened 2025-06-03
fix tests after change in `testthat::expect_equal`

gogonzo:
Looks like `testthat::expect_equal` have changed. It doesn't recognize list sorted in a different way as equal. It also notices the difference between `named list(0)` and `list(0)`

Above is just the assumption, the fact is that some tests doesn't work anymore.


averissimo:
I'm good with whichever of the 3 solutions you prefer:

- Merge this PR _(but first remove the `tests/testthat/helper-all.equal.R`)_
- Remove the `if` clause in the helper
- Remove the helper and register the method to the package

averissimo (approved):
Hahahaha you do love the <span style="color:red;" class="fgColor-danger">**negative**<span>  count.

It looks good

PR #385, opened 2025-08-21
Relax restrictions for indirect link

### LINKED ISSUE #384: [Bug]: Can't find join_keys between indirectly linked datasets
### What happened?

`join_keys` with correct parent-child relationships are not easy to design, especially by the inexperienced developer. Indirectly linked datasets need to have the same parent. IMO this is to strict given that their implicit relationships are "indirect" anyway. Here is an example

Image

```r
jk <- join_keys(
  join_key("customers", keys = "id"),
  join_key("orders", keys = c("id")),
  join_key("products", keys = c("id")),
  join_key("product_components", keys = c("id")),
  join_key("customers", "orders", keys = c(id = "customer_id")),
  join_key("orders", "order_items", keys = c(id = "order_id")),
  join_key("order_items", "products", keys = c(product_id = "id")),  # no common parent
  #join_key("products", "order_items", keys = c(id = "product_id")), # is common parent
  join_key("products", "product_components", keys = c(id = "product_id"))
)

jk["order_items", "product_components"]
# NULL
```

I propose this change to make it easier for app developer as I couldn't figure out what is the problem with above join keys setup myself.

gogonzo:
Closes 
- #384 

For more info and example please check an issue and added test.


R/join_keys-utils.R
```diff
@@ -106,48 +106,46 @@ assert_compatible_keys2 <- function(x, y) {
...
+  }
+  x
+}
 
-        fk <- structure(
-          names(keys_d2_parent)[common_ix_2],
-          names = names(keys_d1_parent)[common_ix_1]
+.search_recursive <- function(x, node_name, parent_name, nodes_visited = node_name) {
+  children_names <- setdiff(names(x[[node_name]]), nodes_visited)
+  nodes_visited <- c(nodes_visited, children_names)
```
osenan: This part is confusing, we reuse the `node_name` if nodes visited are not defined, then nodes visited depend `children_names` and `node_name`. Could it be something more simple like:
```
search_recursive <- function(x, node_name, parent_name, nodes_visited = NULL) {
  if (is.null(nodes_visited))  nodes_visited <- node_name
  children_names <- setdiff(names(x[[node_name]]), nodes_visited)
  nodes_visited <- c(nodes_visited, children_names)
```

R/join_keys-utils.R
```diff
@@ -106,48 +106,46 @@ assert_compatible_keys2 <- function(x, y) {
...
+  for (node_name in names(x)) {
+    x <- .search_recursive(x, node_name)
+  }
+  x
+}
 
-        fk <- structure(
-          names(keys_d2_parent)[common_ix_2],
-          names = names(keys_d1_parent)[common_ix_1]
+.search_recursive <- function(x, node_name, parent_name, nodes_visited = node_name) {
```
osenan: In my opinion, it would be helpful to add internal documentation on this function. At least what is the objective of the function. It is not straightforward to me to understand what it does, besides finding join_keys between two datasets

osenan (approved):
Hi, good changes!

My comment is on function `.search.recursive`. This function is important. I would slightly simplify its beginning and add internal documentation, it can be helpful for the future to debug.

gogonzo:
@osenan I've applied changes in the commented places. 
I've also removed `names.join_keys` which I've recently added. Problem with new `names.join_keys` is that returns topological order while items remain in the same spot in the list. This causes some problems. Specifically, imagine situation when `jk <- list(a = 1, b = 2, c = 3)` and `names(jk)` returns for example `c, a, b`, then:

- `jk[1]` and `jk["c"]` doesn't refer to the same object
- `names(jk)[1] <- "newname"` fails, by replacing `jk[1]` and not `jk["c"]`

Something to think about later. Right now `names(<join_keys>)` returns original list order. I wish this to be ordered topologically which helps with determining merge operations (in `picks`).

gogonzo:
@osenan I need to spend more time on this. `.append_indirect_link` is too slow for larger `join_keys` object:

```r
print(defalt_cdisc_join_keys) # takes a loooooot!
```

osenan (dismissed):
Thanks for the changes, now both documentation and the steps of the function are much easier to understand.

osenan:
@gogonzo I would create a new issue for the performance improvement. We should see if we can do the recursive search with already existing functions from  other packages that are optimized (which would include extra depenencies), or don't use brute force seach if possible

gogonzo:
> I would create a new issue for the performance improvement.

@osenan I won't merge a new feature which will affect all downstream packages. Algorithm is difficult, it is still recursive (not a brute force) but continues when node is re-visited. Revisit on the node is perhaps needed to find "indirect" connection. I've tried to improve performance by skipping "indirect"

I'll come back to this issue later when I have some time off from mdr. Issue I've created is not so crucial as I found more `join_keys` related problems

